PR #3, opened 2025-10-14
patches for Quest browser

cabanier:
Some fixed to make gpu depth work on Quest.
There are still some issues including excessive memory use in the AI samples.


ruofeidu:
Maybe add an option and URL param to allow switching resolutions --- the original resolution may still work for some use cases like coding and inspect elements ;)

cabanier:
> I got a change to try on Quest 3 yesterday. In gemini-icebreakers, opening the same directly not from docs iframe, the mic button only causes the system to hang the first time as it requests permissions. 

We recommend that people ask for permissions before going immmersive. We're aware that this doesn't work great in Quest browser

cabanier:
Thanks! I think if we switch to texture arrays everywhere, the code will become much cleaner.

ruofeidu:
I'm not sure if WebXR folks read my DepthLab paper when designing the Depth Sensing APIs: https://duruofei.com/papers/Du_DepthLab-Real-time3DInteractionWithDepthMapsForMobileAugmentedReality_UIST2020.pdf

I still believe both CPU and GPU data structures should be exposed to developers with different use cases. We are trying hard to patch the existing one... Yet in reality, the performance and human bandwidth are the most challenging blockers ;(

cabanier:
> I'm not sure if WebXR folks read my DepthLab paper when designing the Depth Sensing APIs: https://duruofei.com/papers/Du_DepthLab-Real-time3DInteractionWithDepthMapsForMobileAugmentedReality_UIST2020.pdf

What is missing from the WebXR API?
Note that we also directly support hit testing so the code that is reading back the depth pixels isn't really needed.

> I still believe both CPU and GPU data structures should be exposed to developers with different use cases. We are trying hard to patch the existing one... Yet in reality, the performance and human bandwidth are the most challenging blockers ;(

I don't really see the need for the CPU path. I'm surprised that Android XR still went that way.

PR #8, opened 2025-10-15
Copy measure scene to demos.

dli7319:
Minor updates:
1. Switch to a threejs font instead of uploading roboto.json
2. Add ignoreReticleRaycast to the measuring tape so we don't raycast onto the cylinder as we're measuring.


demos/measure/MeasuringTape.js
```diff
@@ -0,0 +1,100 @@
...
+    if (this.textGeometry) {
+      this.textGeometry.dispose();
+    }
+    const textShapes = this.textFont.generateShapes(this.getLengthText());
+    this.textGeometry = new THREE.ShapeGeometry(textShapes);
+    this.textGeometry.computeBoundingBox();
+    const tempVector = new THREE.Vector3();
+    this.textGeometry.boundingBox.getCenter(tempVector);
+    this.textGeometry.translate(-tempVector.x, -tempVector.y, -tempVector.z);
+    if (this.textMesh) {
```
ruofeidu: nit: let's change it to text view or label view in the next iteration :)

dli7319: The nice thing is that this is self-contained. It works without importing troika.

PR #20, opened 2025-10-19
Clean up unnecessary `material.needsUpdate = true`

dli7319:
Based on [three.js docs](https://threejs.org/docs/#api/en/materials/Material.needsUpdate), needsUpdate signals for the material to be recompiled. Setting uniforms doesn't require recompiling the material.


ruofeidu (approved):
Thanks!

PR #43, opened 2025-11-01
Remove linux from onDesktopUserAgent

dli7319:


dli7319:
Personally, I don't think we should be checking the user agent string at all. We should unify the behavior between devices or check for capabilities only.

ruofeidu:
I agree that we shouldn't, yet have you checked simulator / WebXR runtime are automatically determined without this, 
by checking availability of APIs?

Ideally, by default:

all desktop: simulator only
all phone / Android XR: WebXR only

while developers can request to show two buttons for users to choose from.

If you've checked already in the simulator logic, I'll LGTM.

dli7319:
The current default behavior is desktops always launch simulator and mobile/android XR either show "Enter WebXR" or webxr unsupported. The simulator won't launch on Android XR unless the user enters webxr and then exits.

dli7319:
To clarify, this PR is to allow the simulator to start on linux desktops instead of showing WebXR is not supported.

ruofeidu (approved):

ruofeidu:
Thanks for fixing my bug! I didn't make a good way to determine desktop at the very beginning.

ruofeidu:
This is breaking all demos on Android XR. My bad for not testing before approving.
For future releases, let's have at least two people try out all templates and basic samples :)

PR #44, opened 2025-11-01
Move splats into the splat anchor instead of directly in the scene.

dli7319:
Previously, they had to be a direct child of the scene. Now, it seems to work as a descendant.


dli7319:
Note the splat still lags when dragging. Not sure why.

PR #49, opened 2025-11-06
Fix: sized Description text and graph

jolicaD:
### Changes made
1) Increased font size of the description text
2) Shrunk the graph

image

### Open issue
There is a static dark box in the back of the room that is extraneous, and probably can be removed.


ruofeidu:
Thank you so much!

PR #60, opened 2025-11-25
Remove options.enableGeminiLive()

dli7319:
It's very confusing why we have enableGeminiLive() and enableAI(). This option adds a gemini live text interface to the simulator. It is only used in /demo/aisimulator/.


qxziuan (approved):
LGTM

PR #65, opened 2025-12-02
Add matchDepthView option to DepthOptions to disable depth reprojection

dli7319:


ruofeidu:
LGTM, aligning with https://www.w3.org/TR/webxr-depth-sensing-1/

PR #83, opened 2025-12-13
Use depth view in object detection

dli7319:
Setting matchDepthView to false allows us to use depth without incurring the performance penalty of Chrome reprojecting the depth. Now, we reproject the depth samples when running object detection which should be much faster.

Tested on Moohan only.


dli7319:
@nsalminen @ruofeidu Can you review this again? I made tons of changes. You can now change the simulator depth fov.

PR #96, opened 2025-12-23
Pin litert version in gesture demos

dli7319:


dli7319:
I pinned to 0.2.1 to get things working again.
Eventually we want to update to the latest version.

PR #109, opened 2025-12-31
use options passed into VideoView

conshus:
I could not figure out why 
videoRow.addVideo({
    src: 'https://storage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4',
    autoplay: false
});

would not stop the video in the UI demo (https://xrblocks.github.io/docs/samples/UI/) from autoplaying. Looked though the code and saw that the options weren't being used in the constructor and that the default values were being used no matter what.

This should allow users to pass in options for the video.

Thanks


dli7319:
Thank you.

conshus:
@dli7319 About how often are new releases published? Just incase I spotted anything else.

PR #112, opened 2026-01-04
WebView UI component

conshus:
Happy New Year!
XR Blocks got me inspired to give conference talks again. Last year, I mostly gave workshops. I'm attempting to be able to stream the video out from my Galaxy XR headset to the conference presentation screen to show slides and demos. To show my slides or maybe a website, I wanted to have WebView UI component. I didn't see one, so I made one with the help of Gemini on the Threejs parts.

I basically followed how the VideoView UI component was set up.

I also have a StackBlitz of a running demo for folks to check out: https://stackblitz.com/edit/stackblitz-starters-v5psc3mf?file=index.html (You have to open the preview in a new tab/window because I don't think you can have an iframe in another iframe) It's structured like the UI sample app so it should be easy to just add it.

Let me know what y'all think.

### DIFF
--- a/src/ui/components/WebView.ts
+++ b/src/ui/components/WebView.ts
@@ -1,7 +1,10 @@
 import * as THREE from 'three';
-import { CSS3DRenderer, CSS3DObject } from 'three/addons/renderers/CSS3DRenderer.js';
-import { View } from '../core/View';
-import { ViewOptions } from '../core/ViewOptions';
+import {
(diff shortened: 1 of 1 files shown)


ruofeidu:
Thanks Dwane! Most experiences including depth sensing and gestures will be analogous, but WebView is indeed challenging in XR when linking to a third-party site :( the closest that we can get is html2canvas but misses lots of interactive components.

PR #117, opened 2026-01-15
Add planes to simulator.

dli7319:
Most code written by Gemini.

Adds a CLI to extract the planes from a GLB file and save it as a json file.
At runtime, we can load this json file in addition to the simulator GLB file.

image

### DIFF
--- a/src/world/planes/DetectedPlane.ts
+++ b/src/world/planes/DetectedPlane.ts
@@ -1,51 +1,65 @@
 import * as THREE from 'three';
+import {SimulatorPlane} from './SimulatorPlane';
 
 /**
  * Represents a single detected plane in the XR environment. It's a THREE.Mesh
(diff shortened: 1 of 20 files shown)


src/world/planes/DetectedPlane.ts
```diff
@@ -1,51 +1,65 @@
 import * as THREE from 'three';
+import {SimulatorPlane} from './SimulatorPlane';
 
 /**
  * Represents a single detected plane in the XR environment. It's a THREE.Mesh
  * that also holds metadata about the plane's properties.
  * Note: This requires experimental flag for Chrome.
```
ruofeidu: Note the exact flag and Chrome version.

dli7319: It requires chrome://flags/#openxr-spatial-entities on device. No flags needed on the simulator.

PR #119, opened 2026-01-15
Add on device mesh detection support.

dli7319:
Tested on Quest 3. Not currently mocked on the simulator.

### DIFF
--- a/templates/meshes/main.js
+++ b/templates/meshes/main.js
@@ -1,0 +1,11 @@
+import 'xrblocks/addons/simulator/SimulatorAddons.js';
+
+import * as xb from 'xrblocks';
+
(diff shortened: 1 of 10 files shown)


templates/meshes/main.js
```diff
@@ -0,0 +1,11 @@
+import 'xrblocks/addons/simulator/SimulatorAddons.js';
+
+import * as xb from 'xrblocks';
+
+document.addEventListener('DOMContentLoaded', async function () {
+  const options = new xb.Options();
+  options.reticles.enabled = true;
```
ruofeidu: Reticles should be enabled by default today? It is hard to think of an app without reticles.

dli7319: Yes. It's enabled by default.

PR #130, opened 2026-01-22
Add template parameter for DetectedObject.

dli7319:
This is a breaking change to add types to DetectedObject instead of having arbitrary properties. Additional data properties will be stored in the `data` property.


dli7319:
Another way to do this is to just remove the data parameter and force people to make classes extending DetectedObject.

PR #138, opened 2026-01-28
Mesh detection Sample

dli7319:
Adds a minimal sample to show mesh detection.
Allows the mesh.vertices to be an array of Float32Array to follow https://immersive-web.github.io/real-world-meshing/


ruofeidu (approved):

dli7319:
Added coloring based on the semantic label.

PR #147, opened 2026-02-05
Add stats object to see the FPS.

dli7319:


ruofeidu:
TY!

Consider adding
`xb.options.enableStats();` for quick access?

We had stats object in our first version, but didn't have time to port to the latest one :)

PR #150, opened 2026-02-06
Disable matchDepthView and reticle in ballpit.

dli7319:
Disabling matchDepthView increases FPS from ~35 to ~45 on Galaxy XR. When nothing is in the scene the FPS is around 60 now.


ruofeidu (approved):

ruofeidu:
How about reticle FPS impact?

Indeed we don't need reticles at all in this scene

PR #160, opened 2026-02-13
Support hovering event bubble up or not by returning boolean

qxziuan:
In onHoverEnter, onHoverExit, onHovering, we now support returning booleans. In downstream overrides, if returning true, the even won't bubble up to its parent anymore.


dli7319:
For consistency and conciseness, can you use `&&` instead of nested `if` statements?

dli7319 (approved):

PR #168, opened 2026-02-17
GeminiManager: Reduce jpeg quality

dli7319:
At 1.0 quality, there are ~45ms (~3 frame) stutters as the blob is converted to a ~1MB base64 string.
At 0.8 quality, there are ~14ms (~1 frame) stutters as the blob is converted to a ~200KB base64 string.


dli7319:
Before: https://screenshot.googleplex.com/8xMPRMJvZPZZedA.png
After: https://screenshot.googleplex.com/rzMoT7atgBzxPPd.png

PR #176, opened 2026-02-23
Fix simulator camera getting stuck moving on macOS

salmanmkc:
## Problem

On macOS, the simulator camera can get stuck moving continuously with no way to stop it.

**Steps to reproduce:**
1. Open any template in the simulator (e.g. `templates/0_basic/index.html`)
2. Hold `Command`, then press `S` — the camera starts drifting backward and never stops, even after releasing all keys

This is caused by a macOS browser quirk where `keyup` events are not fired for keys pressed while Command (Meta) is held. The `S` key gets added to the tracked pressed-keys set but is never removed, so the camera moves forever.

This also happens with other movement keys (, `A`, `D`, `Q`, `E`) when combined with `Command`. It doesn't happen with `W` since that will close the tab.

## Fix

Clear all tracked key presses when:
- A `keydown` event has `event.metaKey` set, or the key code is `MetaLeft`/`MetaRight` — prevents the movement key from being registered in the first place
- The window fires a `blur` event — handles app switching, browser dialogs, etc.
- The document fires `visibilitychange` — handles tab switches and system dialogs

## Testing

Tested on a MacBook (macOS, Chrome):
- Held `Command` then pressed `S` — camera no longer gets stuck (previously drifted backward indefinitely)
- Normal movement with `W`/`A`/`S`/`D`/`Q`/`E` still works as expected
- `Shift` to switch simulator modes still works
- `Cmd+Tab` while moving — camera stops correctly when switching apps
- Clean build with `npm run build` (`--failAfterWarnings`), no warnings

### DIFF
--- a/src/simulator/SimulatorControls.ts
+++ b/src/simulator/SimulatorControls.ts
@@ -45,6 +45,7 @@ export class SimulatorControls {
   private _onKeyDown = this.onKeyDown.bind(this);
   private _onKeyUp = this.onKeyUp.bind(this);
   private _onPointerMove = this.onPointerMove.bind(this);
+  private _onBlur = this.onBlur.bind(this);
 
   /**
    * Create the simulator controls.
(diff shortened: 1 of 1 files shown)


dli7319:
Thanks

ruofeidu:
Thank you Salman!

PR #179, opened 2026-02-23
Fix TextView.dispose() failing to remove synccomplete listener

### LINKED ISSUE #178: TextView.dispose() fails to remove synccomplete listener due to .bind() mismatch
## Description

`TextView.dispose()` calls `removeEventListener` with `.bind(this)`, which creates a new function reference each time. Since the listener was originally added with a different `.bind(this)` call, `removeEventListener` silently fails and the `synccomplete` listener is never actually removed.

## Where

In `src/ui/components/TextView.ts`:

**Adding the listener (line ~372):**
```ts
this.textObj.addEventListener('synccomplete', this.onSyncComplete.bind(this));
```

**Removing the listener in `dispose()` (line ~413):**
```ts
this.textObj.removeEventListener('synccomplete', this.onSyncComplete.bind(this));
```

Each `.bind(this)` call creates a new function — these are different references, so `removeEventListener` doesn't match the original and does nothing.

## Impact

The `synccomplete` listener remains attached after `dispose()` is called, which means:
- The callback keeps firing on disposed `TextView` instances
- The disposed `TextView` and its associated Three.js objects can't be garbage collected

## Expected behaviour

`dispose()` should actually remove the listener so the `TextView` can be properly cleaned up.

## Fix

Store the bound function as a class property and use the same reference for both `addEventListener` and `removeEventListener`.

salmanmkc:
Fixes #178

## Problem

`TextView.dispose()` calls `removeEventListener('synccomplete', this.onSyncComplete.bind(this))`, but `.bind(this)` creates a **new function reference** every time it's called. The listener was originally added with a different `.bind(this)` call, so `removeEventListener` doesn't find a match and silently does nothing.

This means the `synccomplete` listener stays attached after `dispose()`, keeping the `TextView` and its Three.js objects alive in memory.

## Fix

Store the bound callback as a class property (`onSyncCompleteBound`) and use the same reference for both `addEventListener` and `removeEventListener`.

## Testing

- Built with `npm run build` (`--failAfterWarnings`) — clean, no warnings
- Ran `npm run lint` — no errors
- Verified the pattern matches how `SimulatorHands` correctly handles bound listeners (using `onHandPoseChangeRequestBound`)

### DIFF
--- a/src/ui/components/TextView.ts
+++ b/src/ui/components/TextView.ts
@@ -146,6 +146,7 @@ export class TextView extends View<TextViewEventMap> {
   /** The total number of lines after text wrapping. */
   lineCount = 0;
 
+  private _onSyncCompleteBound = this.onSyncComplete.bind(this);
   private _initializeTextCalled = false;
   private _text = 'TextView';
   set text(text) {
@@ -372,7 +373,7 @@ export class TextView extends View<TextViewEventMap> {
       this.textObj.addEventListener(
         // @ts-expect-error Missing type in Troika
         'synccomplete',
-        this.onSyncComplete.bind(this)
+        this._onSyncCompleteBound
       );
 
       if (this.imageOverlay) {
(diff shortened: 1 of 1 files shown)


salmanmkc:
> Thanks

you're welcome!

PR #186, opened 2026-03-01
Add new 3DGS walkthrough demo

nsalminen:
Adds a new 3DGS scene walkthrough demo that lets users explore room-scale environments captured with 3D Gaussian Splatting. The demo ships with two scenes derived from the [Zip-NeRF](https://jonbarron.info/zipnerf/) dataset and optimized to be lightweight enough to render smoothly across a range of XR devices. More scenes can be added in the future without significant code changes.

This PR also exports `SparkRendererHolder` from the SDK, which is needed by any app that creates its own `SparkRenderer` and wants the simulator's color space correction (introduced in #37) to apply correctly.

## Features

- **Two room-scale 3DGS scenes** (NYC, Alameda) served from the proprietary assets CDN, based on Zip-NeRF captures
- **Smooth crossfade transitions** between scenes via a 1.5s long pinch or number keys (1, 2)
- **WASD locomotion** in XR via reference space offset
- **Correct color space handling** — creates and registers a `SparkRenderer` so the simulator can toggle `encodeLinear` as introduced in #37
- **`SparkRendererHolder` exported** from `xrblocks` for use in external demos that manage their own `SparkRenderer`
- **Docs page** added to the Advanced Demos section

### DIFF
--- a/demos/3dgs-walkthrough/main.js
+++ b/demos/3dgs-walkthrough/main.js
@@ -39,6 +39,8 @@ const moveDirection = new THREE.Vector3();
  */
 class WalkthroughManager extends xb.Script {
   async init() {
+    this.add(new THREE.HemisphereLight(0xffffff, 0x666666, 3));
+
     // Load all splat meshes in parallel.
     this.splatMeshes = await Promise.all(
       SPLAT_ASSETS.map(async (asset) => {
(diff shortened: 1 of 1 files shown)


ruofeidu:
Thank you Nels! I can test tomorrow.

PR #202, opened 2026-03-22
Fix/camera play error handling + test harness setup

### LINKED ISSUE #196: trouble with the camera
Hi! Awesome project!!! Thank you so much. 
**Description**
One issue I ran into is with the camera. I tried something related to the camera template https://xrblocks.github.io/docs/templates/Camera/, which works on my computer in simulator mode. However, even this template doesn't run when I open it either with my androind phone or with a meta quest 3 headset. The camera is either just black (quest3) or it captures like the first frame after entering and keeps displaying that, but without updating. Occasionally, there appears to be a screen refresh, but the same initial image keeps being displayed, no a continuously evolving camera-view I'm expecting. As said, it works in simulator mode, but I want the real-world camera to do object detection. 

**Steps to reproduce**
Open template https://xrblocks.github.io/docs/templates/Camera/ with phone or vr headset, then move around and observe whether the panel shows the updated camera view.

**Expected behaviour**
The camera view should update as you look around

**Root cause**
?

**Environment**
Android phone (motorola) with chrome browser
Meta quest 3 with internal browser (basic XR works fine).

salmanmkc:
Fixes #196

The camera feed on Android Chrome in WebXR would never render at all, `getUserMedia` fails inside XR sessions, and there was no fallback. This PR implements a full WebXR Raw Camera Access API fallback so the camera actually works, fixes the video playback promise handling, and resolves visual artifacts I noticed once the feed was rendering. I've also added vitest to lock down the stream lifecycle logic with tests.

This took a lot of testing to fix. I tested this on my Android phone — it works in Chrome, Samsung Internet still has the same issue.

Before:
Error on video

After:

Video demo:

https://github.com/user-attachments/assets/6dfa8086-8770-4072-b50f-367a68e0dc24

Video streaming works

### What was happening before
- Loading the camera feed in WebXR on Android Chrome would just fail. `getUserMedia` doesn't work inside an XR session, so the console would swallow `DOMException: play() can only be initiated by a user gesture` and `The play() request was interrupted by a new load request` errors, and the camera stream would never render.
- Upon getting the camera to finally render via the new fallback, I noticed that overlapping UI elements (like `SpatialPanel`) would cause z-fighting/transparency sorting issues, rendering a dark bounding box over the video texture.

### Changes Compared to `main`

1. **WebXR Raw Camera Access API Fallback (`XRDeviceCamera.ts` & `Core.ts`)**
   - This is the main fix. When `getUserMedia` fails during an XR session, the camera now falls back to the WebXR Raw Camera Access API (`getCameraImage`).
   - Added `setRenderer()` so `Core.ts` passes the `THREE.WebGLRenderer` to `XRDeviceCamera` at init time.
   - Added `updateXRCamera(frame)` — called every frame from the render loop in `Core.ts` — which grabs the camera image via `XRWebGLBinding.getCameraImage()`, wraps it in a `THREE.ExternalTexture`, and wires it up as the stream texture.
   - `Core.ts` now requests `camera-access` as an optional WebXR feature and calls `updateXRCamera` each frame when the fallback is active.
   - Added `XRCamera` type declarations in `global.d.ts` for the Raw Camera Access API.

2. **Video Playback & Stream Fixes (`XRDeviceCamera.ts`, `VideoStream.ts`, `VideoView.ts`)**
   - Reworked the `play()` promise handling in `XRDeviceCamera.ts` to catch instead of reject on autoplay policy errors — the video may still autoplay despite the initial rejection.
   - Removed the problematic `video.onerror` reject handler and `video.src = ''` reset that were causing race conditions.
   - Changed `VideoStream.texture` type from `THREE.VideoTexture` to `THREE.Texture` so it can hold the `ExternalTexture` from the camera access fallback.
   - Added a custom `texture.update()` function in `VideoStream` to keep the texture refreshing when `srcObject` changes.
   - `VideoView.load()` now accepts `THREE.Texture` (not just `VideoTexture`), with a new `loadFromTexture()` method for generic textures like `ExternalTexture`.
   - `loadFromStream` now branches between `VideoTexture` and generic `Texture` paths depending on what the stream provides.

3. **Fixed Dark Square Artifacts (`VideoView.ts`)**
   - `transparent: true` was already set on `main`, but without disabling blending it caused z-fighting/transparency sorting bugs (the "dark squares") against UI panels. The actual fix was the three new material properties:
     - `blending: THREE.NoBlending` — strips out alpha blending math, directly fixing the dark bounding box artifacts.
     - `depthWrite: true` (was `false`) — ensures the video plane correctly occludes as an opaque object in the depth buffer.
     - `toneMapped: false` — bypasses global tonemapping so the raw camera feed colors from WebXR aren't artificially dimmed.

4. **Added Vitest**
   - Set up `vitest` with `jsdom` environment in `vitest.config.ts`, `package.json`, and `tsconfig.json`.
   - Added test suites: `XRDeviceCamera.test.ts` and `VideoStream.test.ts`.

### DIFF
--- a/src/ui/components/VideoView.ts
+++ b/src/ui/components/VideoView.ts
@@ -73,8 +73,10 @@ export class VideoView extends View {
 
     const videoGeometry = new THREE.PlaneGeometry(1, 1);
     const videoMaterial = new THREE.MeshBasicMaterial({
-      transparent: false, // Testing if WebXR external texture transparent bug is alpha-blending randomly
-      depthWrite: false,
+      transparent: true,
+      blending: THREE.NoBlending,
+      toneMapped: false,
+      depthWrite: true,
       side: THREE.DoubleSide,
       // `map` will be set based on options.texture or during load
     });


salmanmkc:
> Tested on Galaxy XR and it's working. Interestingly, without this PR, the camera is working already on Galaxy XR but not on phone.

Yeah seems to be problem with phones, glad it worked on Galaxy XR, thanks for testing it so fast and the quick review!

PR #239, opened 2026-04-02
VRM Avatar — Point-to-Walk Companion Demo

Peter-Pater:
## VRM Avatar — Point-to-Walk Companion Demo

Adds a working VRM avatar demo to `demos/vrm-avatar/` using `@pixiv/three-vrm`. Click or pinch a floor location and the avatar walks there, then returns to idle.

**What's included:**
- VRM loading via `GLTFLoader` + `VRMLoaderPlugin` (direct, not `xb.ModelViewer` — see below)
- Mixamo FBX animation retargeting onto the VRM humanoid skeleton with full `MIXAMO_VRM_RIG_MAP`
- Idle/walk crossfade via `AnimationMixer`
- Procedural eye blink via VRM expression manager
- Point-to-walk: raycasts against `xb.core.depth.depthMesh` on device, falls back to y=0 ground plane in simulator
- Spring bone physics active via `vrm.update(delta)`
- Works in the XRBlocks desktop simulator and WebXR

**Why not `xb.ModelViewer`?**
`VRMLoaderPlugin` must be registered on the `GLTFLoader` instance before the load call. `xb.ModelViewer` has no loader injection point, so direct loading is the only viable approach.

**Assets:**
- VRM model loaded from jsDelivr CDN (three-vrm sample model, no local file)
- `animations/Idle.fbx` and `animations/Walking.fbx` — free Mixamo downloads, included in repo


dli7319:
Having some instructions isn't ideal but we can go with that for now if it's the only option.

Have you looked into how https://pixiv.github.io/three-vrm/packages/three-vrm/examples/humanoidAnimation/index.html works?

Do they have it hosted on a CDN somewhere?

ruofeidu (approved):

ruofeidu:
Thank you Xincheng for the contribution and looking into the license!

ruofeidu (approved):

PR #243, opened 2026-04-07
Add option to disable the simulator camera.

dli7319:
In the case that we want to use the simulator but not trigger the simulator camera.


ruofeidu (approved):

ruofeidu (approved):
As long as we don't break https://xrblocks.github.io/docs/templates/Camera/ :)

PR #245, opened 2026-04-10
Mediapipe object detection

samiriff:
Enable local, on-device object detection using MediaPipe, providing an alternative to the cloud-based Gemini detector. To achieve this, the ObjectDetector has been refactored to use a "Backend" strategy, allowing for easier expansion to other detection frameworks in the future.

- Modular Detector Backends: Refactored ObjectDetector.ts to delegate detection logic to specific backend classes (GeminiDetectorBackend and MediaPipeDetectorBackend) using a new BaseDetectorBackend class
- MediaPipe Integration: Added @mediapipe/tasks-vision dependency and implemented MediaPipeDetectorBackend.
- Configuration: Updated ObjectsOptions.ts to include default WASM and model paths for MediaPipe (efficientdet_lite0).


src/world/objects/ObjectDetectorBackend.ts
nsalminen, line 1: It could be worth splitting this file before it grows further, shared backend types/base class in one file, Gemini in its own file, MediaPipe in its own file. Perhaps in a folder specifically for backends. I recommend considering keeping the shared abstractions in `ObjectDetector` and moving concrete backends out.

src/world/objects/ObjectDetectorBackend.ts
```diff
@@ -0,0 +1,434 @@
+import * as THREE from 'three';
+import * as MEDIAPIPE from '@mediapipe/tasks-vision';
```
nsalminen: MediaPipe is imported at module scope here, so the whole object detector path now depends on this package being resolvable even when the active backend is Gemini. I recommend moving MediaPipe into its own backend file and lazy-loading it inside the mediapipe path. 

Otherwise apps that just import `xrblocks` can fail to load unless they also resolve MediaPipe, even if they never use object detection

src/camera/CameraUtils.ts
```diff
@@ -250,3 +250,51 @@ export async function cropImage(base64Image: string, boundingBox: THREE.Box2) {
...
+
+/**
+ * Asynchronously crops an ImageData object using a THREE.Box2 bounding box.
+ * This function draws a specified portion of the ImageData to a canvas,
+ * and then returns the canvas content as a new base64 string.
+ * @param imageData - The source ImageData object.
+ * @param boundingBox - The bounding box with relative coordinates (0-1) for cropping.
+ * @returns A promise that resolves with the base64 string of the cropped image.
+ */
+export async function cropImageData(
```
nsalminen: There is some code duplication with `cropImage`. Could we extract a helper function to keep this file concise?

src/world/objects/ObjectDetectorBackend.ts
```diff
@@ -0,0 +1,434 @@
...
+
+  protected async detect(
+    snapshot: CameraSnapshot
+  ): Promise<NormalizedDetectedObject<T>[]> {
+    const mediapipeOptions =
+      this.context.options.objects.backendConfig.mediapipe;
+    const vision = await MEDIAPIPE.FilesetResolver.forVisionTasks(
+      mediapipeOptions.wasmFilesUrl
+    );
+    const objectDetector = await MEDIAPIPE.ObjectDetector.createFromOptions(
```
nsalminen: I’d recommend reusing a cached detector instance instead of creating a new one for each run, since repeated detections will otherwise pay the full initialization cost every time.

src/world/objects/ObjectsOptions.ts
```diff
@@ -38,8 +38,14 @@ export class ObjectsOptions {
         },
       },
     },
-    /** Placeholder for a future MediaPipe backend configuration. */
-    mediapipe: {},
+    /** Configuration for MediaPipe backend. */
+    mediapipe: {
+      wasmFilesUrl:
+        'https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision@latest/wasm',
```
nsalminen: I’d recommend pinning the WASM URL to the same MediaPipe version as the package import, since using @latest here can lead to JS/WASM version drift.

src/world/objects/ObjectDetectorBackend.ts
```diff
@@ -0,0 +1,434 @@
...
+    cameraParametersSnapshot: CameraParametersSnapshot
+  ): Promise<DetectedObject<T>[]> {
+    if (!this.isAvailable()) {
+      return [];
+    }
+
+    const snapshot = await this.getSnapshot();
+    if (!snapshot) return [];
+
+    const normalizedDetections = await this.detect(snapshot);
```
nsalminen: I’d recommend wrapping this shared detection call in error handling, so backend init or inference failures fail gracefully instead of rejecting the whole detection flow.

src/world/objects/ObjectDetectorBackend.ts
```diff
@@ -0,0 +1,434 @@
...
+}
+
+/**
+ * Object detector backend implementation using MediaPipe's Object Detector.
+ * Runs locally on the device.
+ *
+ * T - The type of additional data associated with the detected object (not used currently).
+ */
+export class MediaPipeDetectorBackend<T> extends BaseDetectorBackend<T> {
+  protected isAvailable(): boolean {
```
nsalminen: I’d recommend making this availability check more defensive, since MediaPipe availability depends on more than code path selection, like module load, WASM fetch, model fetch etc.

nsalminen (changes requested):
Thank you so much @samiriff! Very cool! I ran your code in the simulator and overall it works well and fast. Have you tested this on an Android XR device yet? Otherwise perhaps @dli7319 or @ruofeidu could help with that as I don't have any at the moment.

I do sometimes get some odd detections or no detections at all, but I guess this may have to do with mediapipe rather than your code. If you do have ideas on how to test whether everything is working correctly (e.g., running it manually with a cached screenshot and a simple mediapipe demo), then that'd be worth checking I think.

I left some comments in your code with some suggestions.

Image

src/world/objects/ObjectDetectorBackend.ts
dli7319: Right now we typically dynamically load the library since there's no universal path to dynamically load individual xrblocks files as they get rolled up together unless they're in addons.

src/camera/CameraUtils.ts
samiriff: Done.

src/world/objects/ObjectDetectorBackend.ts
samiriff: Sounds good. Done.

src/world/objects/ObjectsOptions.ts
samiriff: Done.

src/world/objects/ObjectDetectorBackend.ts
samiriff: Done.

src/world/objects/ObjectDetectorBackend.ts
samiriff: Makes sense. I have separated out the backend detectors in a new "backends" directory, and also merged the contents of `ObjectDetectorBackend.ts` with `ObjectDetector.ts`.

src/world/objects/ObjectDetectorBackend.ts
samiriff: I have conditionally imported the GeminiDetectorBackend and MediaPipeDetectorBackend by adding `inlineDynamicImports: true` to the `rollup_config.js`. It seems to work in the simulator locally, so I hope this is fine.

src/world/objects/ObjectDetectorBackend.ts
samiriff: Done.

samiriff:
Thanks for your comments, @nsalminen! I have tried to incorporate your suggestions wherever I could. 

1. Regarding the odd detections with Mediapipe, I think that might be because I was using the low-latency-low-accuracy model (https://ai.google.dev/edge/mediapipe/solutions/vision/object_detector#efficientdet-lite0_model_recommended). I've updated the config option to use a high-accuracy-high-latency model (https://ai.google.dev/edge/mediapipe/solutions/vision/object_detector#efficientdet-lite2_model), which works better but is not as good as Gemini. Hope this is fine for now.

2. Regarding physical device testing, I will try to test out this code on a physical XR device in my office as soon as possible but I'll reach out in case I run into issues.  

3. Regarding tests, I can try to add a complete set of unit tests for both the Gemini and MediaPipe backends in a follow-up PR. I'll start looking into this immediately.

samiriff:
@nsalminen - Good news regarding point 2. I was able to test MediaPipe object detection on a physical XR device and it worked successfully (http://screen/9aZLeBm9hcSMGbj).

src/world/objects/ObjectDetectorBackend.ts
dli7319: I don't think that's what we want.
The goal is to not have `import * as MEDIAPIPE from '@mediapipe/tasks-vision';` at the top of the built `xrblocks.js` file. Otherwise all apps which use xrblocks will need to also include `@mediapipe/tasks-vision`.

The two ways we've been doing it so far is:
1. Make a dynamic import which gets called only if the option to use it is set, used for @google/genai: https://github.com/google/xrblocks/blob/5bf6b0d6205763f8c2019210bc47a5c05458ceb0/src/ai/Gemini.ts#L17
2. Make the module itself an option to be passed in, used for RAPIER: https://github.com/google/xrblocks/blob/5bf6b0d6205763f8c2019210bc47a5c05458ceb0/src/physics/Physics.ts#L29

samiriff: Thanks, David! I have updated the imports to follow the approach similar to the one for Gemini. Please let me know if this is fine.

dli7319 (approved):

nsalminen (approved):

PR #262, opened 2026-05-08
Add gamepad support to simulator

salmanmkc:
Adds gamepad support to the simulator so people on PC who don't want to switch to keyboard/mouse can use an Xbox/PS controller instead. WebXR already supports controllers, but the desktop simulator preview didn't.

Sticks move and look, triggers go up/down, RB toggles the UI, Start opens a settings panel for rebinding buttons, D-pad cycles hand poses. In Controller mode the stick also moves the active hand. Bindings are stored in localStorage so they persist.

When the settings panel is open it grabs gamepad input so the camera doesn't fly around while you're picking menu items.

Tested manually with an Xbox controller. New unit tests cover the bindings store and rising-edge button timing.

Demo: https://youtu.be/RgJJfYdkvJM

### DIFF
--- a/src/addons/simulator/SimulatorAddons.ts
+++ b/src/addons/simulator/SimulatorAddons.ts
@@ -1,3 +1,5 @@
 import './instructions/SimulatorInstructions.js';
+import './ui/GamepadSettingsPanel.js';
+import './ui/GamepadToast.js';
 import './ui/HandPosePanel.js';
 import './ui/ModeIndicator.js';
--- a/src/addons/simulator/ui/GamepadSettingsPanel.ts
+++ b/src/addons/simulator/ui/GamepadSettingsPanel.ts
@@ -15,6 +15,12 @@ const ACTION_LABELS: Record<xb.GamepadAction, string> = {
   openSettings: 'Open Settings',
 };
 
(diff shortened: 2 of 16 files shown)


ruofeidu:
Thank you Salman for the valuable contribution! This is indeed very useful!

David: would you give it a review pass?

In the long run, we need to design a settings menu for simulator to debug controller, hands, depth, accessibility etc. 
For now, the mapping of the controller to the simulator works.

For hands, another direction is to add webcam-based simulation of hands.

dli7319:
Will test this after work since I don't have a gamepad handy in the office.

salmanmkc:
> Thank you Salman for the valuable contribution! This is indeed very useful!
> 
> David: would you give it a review pass?
> 
> In the long run, we need to design a settings menu for simulator to debug controller, hands, depth, accessibility etc. For now, the mapping of the controller to the simulator works.
> 
> For hands, another direction is to add webcam-based simulation of hands.

That's a cool idea! 


> Will test this after work since I don't have a gamepad handy in the office.

Awesome thanks!

dli7319:
Just tested and it's working well. Great job!
The UI is a nice touch.
image

dli7319 (approved):

PR #261, opened 2026-05-06
Add shader portals demo

salmanmkc:
Adds a portal demo alongside the other demos in `demos/` (ballpit, rain, xrpoet, etc).

5 shader-disc portals in an arc — underwater, forest, cosmic, lava, and cyberpunk. Each one is a self-contained procedural world. You can pick a portal up and place it on a surface using the depth mesh.

Everything is single-pass fragment shaders, no render targets or stencils. Adding more worlds is just a new file in `scenes/` and a `SCENES` entry.

`npm run serve` → `http://localhost:8080/demos/cosmicportal/`

PR #265, opened 2026-05-09
Walk-in portal immersives

salmanmkc:
Follow-up to #261. Same portal demo, but you can now walk *into* a portal and end up inside the world.

Each scene has a matching immersive sphere with true 3D raymarching, not the 2D parallax disc. Walk through the disc and the world swaps to the immersive shader; walk back out the other side and you're back in the lobby. Bidirectional, so the entry isn't blocked behind you.

A few other things picked up along the way:

- per-eye stereo parallax on the disc (was flat before)
- the lobby pauses non-active portals while you're inside one
- forest immersive mirrors the cinematics from the disc (owl, comet, lightning, aurora) so going in doesn't feel like a different scene
- cosmic ring is now a drifting rainbow, the rest still use their 2-color bands
- lava: cone-shaped volcanoes with proper 3D bombs, dropped a 2D explosion artifact from the preview
- cyberpunk: night park with neon billboards
- underwater: warmer sand + a school of Dory tang & clownfish

Most of the perf cost is the immersive shader. The exit-snapshot RT is now 256² and runs every other frame, and the hot per-frame allocations in the gallery are gone. Still GPU-bound on M1 Chrome but should be fine on the Galaxy XR target.

`npm run serve` → `http://localhost:8080/demos/portals/`

Demo: https://youtu.be/TMl-VkBOv50

### DIFF
--- a/demos/portals/PortalGalleryScene.js
+++ b/demos/portals/PortalGalleryScene.js
@@ -52,7 +52,7 @@ const TRANSITION_COOLDOWN_MS = 600; // Lock new transitions briefly after each
 export class PortalGalleryScene extends xb.Script {
   portals = [];
   labels = [];
-  immersives = []; // index → ImmersiveInstance | null
+  immersives = []; // index -> ImmersiveInstance | null
   clock = new THREE.Clock();
   _held = null;
   _activeIndex = -1; // Portal index user has walked into, or -1.
@@ -167,9 +167,9 @@ export class PortalGalleryScene extends xb.Script {
 
     // Immersive worlds (one per scene that supports walk-in).
     for (let i = 0; i < SCENES.length; i++) {
-      const Cls = IMMERSIVE_BY_NAME[SCENES[i].name];
-      if (Cls) {
-        const inst = new Cls();
+      const ImmersiveClass = IMMERSIVE_BY_NAME[SCENES[i].name];
+      if (ImmersiveClass) {
+        const inst = new ImmersiveClass();
         this.add(inst);
         this.immersives.push(inst);
       } else {
--- a/demos/portals/PortalGalleryScene.js
+++ b/demos/portals/PortalGalleryScene.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three';
 import * as xb from 'xrblocks';
 
+import {CosmicImmersive} from './CosmicImmersive.js';
 import {Portal} from './Portal.js';
 import {CosmicScene} from './scenes/CosmicScene.js';
 import {CyberpunkScene} from './scenes/CyberpunkScene.js';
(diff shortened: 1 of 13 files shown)


ruofeidu (approved):

salmanmkc:
Yay thanks for the quick review!

demos/portals/PortalGalleryScene.js
```diff
@@ -28,12 +47,32 @@ const ARC_SPAN = Math.PI * 0.85; // total arc span (~150°)
  *   2. Click anywhere on the depth mesh to drop it there, snapped to the
  *      surface normal.
  *   3. Click the held portal again (or click empty space twice) to cancel.
+ *   4. Walk through the Cosmic portal to enter immersive space mode.
  */
 export class PortalGalleryScene extends xb.Script {
   portals = [];
   labels = [];
+  immersives = []; // index → ImmersiveInstance | null
```
ruofeidu: probably use -> and avoid unicode for comments?

demos/portals/PortalGalleryScene.js
```diff
@@ -80,9 +119,69 @@ export class PortalGalleryScene extends xb.Script {
...
+      fadeMat
+    );
+    this._fadeSphere.renderOrder = 9999;
+    this._fadeSphere.frustumCulled = false;
+    this._fadeSphere.raycast = () => {}; // Don't intercept clicks.
+    this.add(this._fadeSphere);
+
+    // Immersive worlds (one per scene that supports walk-in).
+    for (let i = 0; i < SCENES.length; i++) {
+      const Cls = IMMERSIVE_BY_NAME[SCENES[i].name];
```
ruofeidu: nit: more descriptive name for Cls

demos/portals/scenes/LavaScene.js
```diff
@@ -48,19 +205,103 @@ export const LavaScene = {
...
+      float fogF = smoothstep(8.0, 80.0, opaqueT);
+      col = mix(opaqueCol, col, fogF * 0.25);
+    }
+
+    // ---- Animated ash plume + lava spurts above main volcano apex ----
+    {
+      vec3 apexW = vec3(2.0, 4.4, -16.0);
+      // Sample 14 horizontal slices from the crater upward; each slice tests
+      // distance from the vertical column through the apex.
+      for (int i = 0; i < 14; i++) {
```
ruofeidu: nit: avoid magic number, use kNumSlices

demos/portals/PortalGalleryScene.js
salmanmkc: good point, thanks!

demos/portals/PortalGalleryScene.js
salmanmkc: also good point, thanks!

demos/portals/scenes/LavaScene.js
salmanmkc: thanks again, updated!

salmanmkc:
Thanks for the feedback, just addressed it!
Should i be putting this now in a sample you think?

ruofeidu:
I think you can list it as an advanced demo now. Thanks Salman!

PR #267, opened 2026-05-10
Add Portal Gallery to docs samples

salmanmkc:
Promoting the portals demo to Advanced Demos (#265), comment [here](https://github.com/google/xrblocks/pull/265#issuecomment-4411247784).

Same pattern as BalloonPop and 3DGS Walkthrough mdx page wrapping it in SamplesIFrame, plus a sidebars.ts entry.

Looks like this:
Docs for portals


dli7319:
TY!

PR #268, opened 2026-05-10
world_companion demo + startGeminiLive captureMode flag (multi-item placeLabel, embedding match)

salmanmkc:
been wanting a voice + vision companion in xr for a while. start it, talk to it, it sees what you see and can drop markers on stuff.

demo's in `demos/world_companion`. it runs on the existing `startGeminiLive` in GeminiManager, which now takes a `captureMode: 'screenshot' | 'camera'` flag, screenshot streams the rendered scene (optionally over the camera image), camera streams raw passthrough frames. default stays `camera` so existing callers keep their old behaviour; the companion passes `camera` since it reasons about the real room. tool calls + transcription come back via the manager's events, so nothing new on the world api.

there's a `placeLabel` tool with three styles (dot, arrow, pulse) so it can pick how to highlight, arrow if you ask it to find something, pulse for tiny stuff, dot otherwise. uses `world.objects.runDetection` so markers stick to the real object via depth, not to wherever your head was when the tool fired, it will do this on the desktop simulator though. a `lookCloser` tool answers "what am i pointing at" off the reticle. small spatial panel for start/stop/clear so it's actually usable once you're in immersive.

you can ask for several things in one go ("label the couch, tv, and coffee table") and the tool takes an `items[]` array so they all get placed in a single call, each with its own style. labels billboard back at the camera so they stay readable when you walk around.

detector labels and what you say don't always match, television vs tv, pendant light vs floor lamp, picture vs painting. it matches by meaning through gemini's embedContent api (SEMANTIC_SIMILARITY task type + cosine), reducing to the core noun first so modifiers like "green chair" don't tank the match, and dedupes markers across rephrasings. the embed cache is per-page so it's basically free after the first call.

Try launch the demo and say for example "place an arrow on my water bottle".

I have my thoughts on updating states of objects it has seen, to update later, however for now this seems ok.

Gemini will be able to talk and see screens afaik in Android XR, however this will allow interaction in the real world + gemini live.

I will see if I can get a demo recorded for this. This is open to lots of feedback though, since this is just a very rough version.

Edit, here's a demo! https://youtu.be/-5s_aV6eV_A

I may have to just add key input in again but will double check later when I'm home


salmanmkc:
Turns out I hit rate limits of 20 object detections per day when I checked logs, I for some reason though it was broken

src/camera/QuestCameraParams.ts
```diff
@@ -0,0 +1,59 @@
+import * as THREE from 'three';
+import {intrinsicsToProjectionMatrix} from './CameraParameterUtils';
+
+// Approximate intrinsics for the Meta Quest 3 front passthrough RGB camera
```
dli7319, line 4: Thanks for estimating this.
Can we name all the constants `QUEST_3_XXX` if we're not sure if quest 3s is different?

salmanmkc: Yup good idea!

salmanmkc:
btw poet can now use world.lookingAt() instead since it's a primitive now in this pr

ruofeidu:
Thanks Salman! I'll test it in Android XR in the afternoon and go through the syntax.

ruofeidu:
Hi Salman, 

Do you have an Android XR device to try?

1. The arrow and the depth query of the object seems mismatching. Nels has an amazing demo here (https://xrblocks.github.io/docs/samples/Gemini-XRObject/) that uses average depth to place where it is by a long pinch gesture. (yes we need a panel to prompt user what to do as well in this demo)
2. Prompt the user with what to ask on top of the panel: E.g., try speaking "place an arrow on my laptop". See the Gemini Icebreakers demo: https://xrblocks.github.io/docs/samples/Gemini-Icebreakers/
3. In XR, only microphone, stop, and delete buttons were shown, I did see some 2D UIs after exiting XR with the transcription --- is this by design?

ruofeidu:
4. You need to request microphone permission before entering XR, wait a bit for David's PR to fix that :)

ruofeidu:
With https://xrblocks.github.io/docs/samples/Gemini-XRObject/, I'm not sure if this demo adds a new capabilities to XR Blocks, or would you correct me if I was wrong. Indeed there is a misalignment with Galaxy XR and Quest and I hope we can reach similar outcome eventually.

salmanmkc:
> Hi Salman,
> 
> Do you have an Android XR device to try?
> 
> 1. The arrow and the depth query of the object seems mismatching. Nels has an amazing demo here (https://xrblocks.github.io/docs/samples/Gemini-XRObject/) that uses average depth to place where it is by a long pinch gesture. (yes we need a panel to prompt user what to do as well in this demo)
> 2. Prompt the user with what to ask on top of the panel: E.g., try speaking "place an arrow on my laptop". See the Gemini Icebreakers demo: https://xrblocks.github.io/docs/samples/Gemini-Icebreakers/
> 3. In XR, only microphone, stop, and delete buttons were shown, I did see some 2D UIs after exiting XR with the transcription --- is this by design?

Ah no I don't have an Android XR device unfortunately, can't order one in the UK 😢 hoping that will change tomorrow

ruofeidu:
I double checked the demo, the arrows were placed on the same distance regardless how far away I'm holding an object on Android XR --- maybe double check the https://xrblocks.github.io/docs/samples/Gemini-XRObject/ for existing APIs.

I'll convert to draft now.

demos/world_companion/index.html
```diff
@@ -0,0 +1,692 @@
...
+                  itemStyle
+                );
+                placed.push({
+                  placed: match.label,
+                  requested: item.text,
+                  style: itemStyle,
+                  anchored: true,
+                });
+              } else {
+                placeMarker(fallbackPosition(i), item.text, itemStyle);
```
salmanmkc: @ruofeidu I think you're hitting the fallback here? Are you rate limited by chance or something? Is it just for arrows? Since this happens when it fails to detect or if rate limited normally.

salmanmkc:
> With https://xrblocks.github.io/docs/samples/Gemini-XRObject/, I'm not sure if this demo adds a new capabilities to XR Blocks, or would you correct me if I was wrong. Indeed there is a misalignment with Galaxy XR and Quest and I hope we can reach similar outcome eventually.

yeah good question, the overlap is real but i think the pitch is different:

 - gemini-xrobject is one-shot "tell me about this thing" — long-pinch → detect → tap → ask
 - world_companion is the opposite, gemini live is always listening + seeing through the camera and it decides when to mark something. so you can just say "find my keys" or "what's that thing on the shelf" without pinching first and it drops a marker mid-convo. the marker styles (arrow / pulse / dot) are there so the model picks how to highlight, not 
always-a-sphere

the new sdk bits (world.streamScene + world.lookingAt) were added to make that loop possible, opening a live session with the camera streaming + tool calls coming back is what enables the continuous thing. gemini-xrobject doesn't need either since it's one-shot.

ruofeidu:
Great explanation, feel free to add a README to highlight the difference :)

I think you can safely use our simulator to debug it... once working in the simulator it is likely it works in Android XR!

The arrow doesn't really point to the object now.

Screenshot 2026-05-11 at 6 16 09 PM

salmanmkc:
wondering if I should use gemini embedContent API as that would help with similarities, vector db is overkill and ya would not be good to make my own local one

salmanmkc:
Nice I think it's at a finalized state now ready for review again, I've added gemini embedding embeddings now so similar words work based on vector difference, it's done on cloud so not local but since we're already doing api calls I think no problem!

src/world/objects/ObjectDetectorBackend.ts
```diff
@@ -206,10 +206,15 @@ export abstract class BaseDetectorBackend<T> {
         .slice(0, 19)
         .replace('T', '_')
         .replace(/:/g, '-');
-      const link = document.createElement('a');
```
dli7319, line 209: Can you revert this?

src/world/World.ts
```diff
@@ -206,4 +220,244 @@ export class World extends Script {
...
+   *
+   * @param prompt - The system instruction / persona for the session.
+   * @param options - Frame rate, image quality, optional tools, and
+   *     callbacks.
+   * @returns A handle exposing `stop()` and `isActive()`. The caller is
+   *     responsible for calling `stop()`.
+   * @throws If no AI is registered, the active model isn't Live-capable, or
+   *     no XRDeviceCamera is registered.
+   */
+  async streamScene(
```
dli7319: Can we refactor this into its own file, e.g. `world/GeminiStreaming.ts`?

dli7319 (commented):
I'm not quite sure if `world.streamScene` is something that belongs in the SDK directly or in `addons/`. @ruofeidu was this in your plans?

We already have https://github.com/google/xrblocks/blob/main/src/addons/ai/GeminiManager.ts which is very similar and doesn't abstract everything into a single function call.

src/world/objects/ObjectDetectorBackend.ts
salmanmkc: yes will do

src/world/World.ts
salmanmkc: I guess does this depend on if it's in SDK or add ons, thoughts @ruofeidu?

ruofeidu:
> I'm not quite sure if `world.streamScene` is something that belongs in the SDK directly or in `addons/`. @ruofeidu was this in your plans?
> 
> We already have https://github.com/google/xrblocks/blob/main/src/addons/ai/GeminiManager.ts which is very similar and doesn't abstract everything into a single function call.

No, this is not my plan. It sounds like an addon feature for Salman's Netblocks.

ruofeidu:
Sorry for the delayed review, can we resolve the conflicts here?

dli7319:
I'd prefer if the functionality of `streamScene` could be merged into https://github.com/google/xrblocks/blob/main/src/addons/ai/GeminiManager.ts instead of being in `world`.

And `lookingAt` seems like just an alias for `getReticleTarget` and has nothing to do with world sensing.

src/addons/ai/GeminiManager.ts
```diff
@@ -389,8 +389,239 @@ export class GeminiManager extends xb.Script<GeminiManagerEventMap> {
...
+   * `streamToAI: true` before this if you want the model to hear the user.
+   *
+   * @param prompt - The system instruction / persona for the session.
+   * @param options - Frame rate, image quality, optional tools, and callbacks.
+   * @returns A handle exposing `stop()` and `isActive()`. The caller is
+   *     responsible for calling `stop()`.
+   * @throws If no AI is available, the active model isn't Live-capable, or no
+   *     XRDeviceCamera is available.
+   */
+  async streamScene(
```
dli7319: This isn't cleanly integrated.
We should modify `startGeminiLive()` or the constructor to configure whether we want screenshots  (existing) or camera snapshots (what you have now).
And everything in `StreamSceneOptions` should be passed in either `startGeminiLive()` or the constructor.
This way we don't need a new `streamScene` entrypoint.

src/world/objects/ObjectDetectorBackend.ts
dli7319: Remember to remove this when you're done.

src/camera/CameraUtils.ts
```diff
@@ -22,6 +26,10 @@ export const DEVICE_CAMERA_PARAMETERS: {[key: string]: DeviceCameraParameters} =
       projectionMatrix: MOOHAN_PROJECTION_MATRIX,
       getCameraPose: getMoohanCameraPose,
     },
+    quest: {
```
dli7319: quest3 or quest_3

src/world/objects/ObjectDetector.ts
```diff
@@ -106,6 +112,14 @@ export class ObjectDetector extends Script {
     this.camera = camera;
     this.renderer = renderer;
 
+    if (
+      this.targetDevice === 'galaxyxr' &&
+      typeof navigator !== 'undefined' &&
+      /OculusBrowser|Quest/i.test(navigator.userAgent)
+    ) {
+      this.targetDevice = 'quest';
```
dli7319: `quest3` or `quest_3`

demos/world_companion/index.html
```diff
@@ -0,0 +1,958 @@
...
+      content="width=device-width, initial-scale=1.0, user-scalable=no"
+    />
+    <link type="text/css" rel="stylesheet" href="../../samples/main.css" />
+    <script>
+      window.litDisableBundleWarning = true;
+    </script>
+    <script type="importmap">
+      {
+        "imports": {
+          "three": "https://cdn.jsdelivr.net/npm/three@0.182.0/build/three.module.js",
```
dli7319: Please update to 184

dli7319 (changes requested):

dli7319:
It looks like most of the changes to GeminiManager disappeared? Is that intentional?

salmanmkc:
> It looks like most of the changes to GeminiManager disappeared? Is that intentional?

yeah that was on purpose. your earlier comment said we don't need a new streamScene entrypoint, so i deleted it and moved everything onto the existing startGeminiLive. only thing i added to GeminiManager is optional cameraWidth/cameraHeight for downscaling the camera frames.

did I misunderstand the approach you had? also did you want the screenshot-vs-camera choice as an explicit mode flag rather than just the dimensions?

dli7319:
The screenshot sends frames of virtual content (with the camera image optionally underneath). I thought your demo required sending camera images.

dli7319:
Apologies for not being clear with my comments.

My thought was to merge the capabilities of your `streamScene` into `startGeminiLive` so an explicit flag decides whether to stream screenshots or stream camera frames. The rest of `GeminiManager` already handles the callbacks and the microphone so those would be redundant.

dli7319:
Your demo video is super cool btw! We're looking forward to merging it!
We can probably delete demos/aisimulator once your demo is merged.

salmanmkc:
> The screenshot sends frames of virtual content (with the camera image optionally underneath). I thought your demo required sending camera images.

ah I see, yea I understand what you meannow

> Apologies for not being clear with my comments.
My thought was to merge the capabilities of your streamScene into startGeminiLive so an explicit flag decides whether to stream screenshots or stream camera frames. The rest of GeminiManager already handles the callbacks and the microphone so those would be redundant.

thank you that makes sense, one thing on defaults: i kept `captureMode: 'camera'` as the default so existing callers of `startGeminiLive` (7_ai_live etc.) keep streaming the passthrough camera like before. screenshot is opt-in. world_companion passes `'camera'` explicitly anyway. lmk if you'd rather screenshot be the default. I've got it setup like this:  `startGeminiLive` now takes `captureMode: 'screenshot' | 'camera'`. screenshot streams the rendered scene via `core.screenshotSynthesizer` (with the camera underneath when `overlayOnCamera` is set), camera streams raw `xrDeviceCamera` frames. the world_companion demo passes `'camera'`

> Your demo video is super cool btw! We're looking forward to merging it! We can probably delete demos/aisimulator once your demo is merged.

thanks! and yeah, makes sense to drop demos/aisimulator once this lands.

demos/world_companion/index.html
```diff
@@ -0,0 +1,1133 @@
...
+              return best;
+            };
+            // Prefer an unlabeled object; fall back to any match (e.g. when the
+            // user re-references the one thing already marked).
+            const findMatch = (text) =>
+              findMatchIn(
+                text,
+                detected.filter((d) => !isLabeled(d))
+              ) || findMatchIn(text, detected);
+            const MAX_ANCHOR_DIST = 8;
```
nsalminen: This could be a good config variable for easy updating.

demos/world_companion/index.html
```diff
@@ -0,0 +1,1133 @@
...
+      }
+
+      function clearLabels() {
+        while (labels.children.length) {
+          const c = labels.children.pop();
+          c.traverse((n) => {
+            n.geometry?.dispose();
+            n.material?.dispose();
+          });
+          labels.remove(c);
```
nsalminen: `pop()` should already remove `c` here, so possibly unnecessary to call `remove()`.

demos/world_companion/index.html
```diff
@@ -0,0 +1,1133 @@
...
+        transcriptEl.innerHTML = '';
+        try {
+          // Clear any leftover labels from prior session to avoid stacking.
+          while (labels.children.length) {
+            const c = labels.children.pop();
+            c.traverse?.((n) => {
+              n.geometry?.dispose?.();
+              n.material?.dispose?.();
+            });
+            labels.remove(c);
```
nsalminen: `pop()` should already remove `c` here, so possibly unnecessary to call `remove()`.

src/camera/QuestCameraParams.ts
```diff
@@ -0,0 +1,62 @@
...
+  QUEST_3_CAMERA_POSE_IN_RIGHT_CAMERA_ROTATION,
+  QUEST_3_CAMERA_POSE_IN_RIGHT_CAMERA_SCALE
+);
+
+export function getQuestCameraPose(
+  _camera: THREE.Camera,
+  xrCameras: THREE.WebXRArrayCamera,
+  target: THREE.Matrix4
+) {
+  target.compose(
```
nsalminen: Could directly reference `QUEST_3_CAMERA_POSE_IN_RIGHT_CAMERA` instead of recomposing.

src/addons/ai/GeminiManager.ts
```diff
@@ -226,6 +304,11 @@ export class GeminiManager extends xb.Script<GeminiManagerEventMap> {
     if (!this.audioContext) {
       this.audioContext = new AudioContext({sampleRate: 24000});
```
nsalminen: `audioContext` is now reused for audio playback and capture. The sample rate that is used depends on which code path initializes it first. `CoreSound` uses separate contexts (in `AudioListener` and `AudioPlayer`, which could also be applied in `GeminiManager`.

My preference is to completely delegate Web Audio operations in `GeminiManager` to `CoreSound` instead so we don't reimplement low-level audio capture and playback.

nsalminen (changes requested):
Very cool demo and great addition to `GeminiManager`! I left some comments on possible optimizations.

src/addons/ai/GeminiManager.ts
dli7319: This was already there before this PR, but would be nice to get fixed.

src/camera/QuestCameraParams.ts
salmanmkc: thank you, done!

src/addons/ai/GeminiManager.ts
salmanmkc: thank you! have done this now, removed AudioCaptureProcessorCode

demos/world_companion/index.html
salmanmkc: thank you! good point, removed it

demos/world_companion/index.html
salmanmkc: thanks again, i've removed it now

demos/world_companion/index.html
salmanmkc: thanks! done

ruofeidu:
WebXR working group does not really want to expose device type while we need to determine device type for camera intrinsics.

@nsalminen @dli7319 : what do you think of this in the long run or shall we follow Salmon's approach here?

Maybe allow users to choose device from a dropdown box before "Enter XR"? Auto determine based on some heuristics or userAgentData?

dli7319:
> WebXR working group does not really want to expose device type while we need to determine device type for camera intrinsics.
> 
> @nsalminen @dli7319 : what do you think of this in the long run or shall we follow Salmon's approach here?
> 
> Maybe allow users to choose device from a dropdown box before "Enter XR"? Auto determine based on some heuristics or userAgentData?

I don't know of any heuristics that work well. User agent is the same between Galaxy XR and Aura.

Ultimately we need an API to get the intrinsics.

salmanmkc:
agreed, the UA thing is a blocker if it can't split Galaxy XR from Aura since they're on the same Android XR user agent. 

So kinda just realised webxr already gives us the real numbers? the raw camera access spec puts the camera's intrinsics on `XRView.projectionMatrix` and its pose on `XRView.transform`. and we're already holding that `view`, `XRDeviceCamera.updateXRCamera` walks `pose.views` and reads the camera image off `view.camera`, it just doesn't read those other two fields.

so what if we read them off the view we already have instead of the table? then the device tells us its own camera math, whatever device it is. the table stays as the fallback for the getUserMedia path (older Quest, no XRView) and the simulator, and `targetDevice` stays as a manual override. honestly not 100% sure it lines up cleanly on every device but it feels like the right direction.

change is small since we're already in that loop. in `updateXRCamera`, next to the width/height we already read:

```ts
// view.projectionMatrix = camera intrinsics, view.transform = camera pose,
// both in the same reference space we use for getViewerPose (and that three
// uses for the scene), so no extra device profile needed.
const xrView = view as XRView & {camera?: XRCamera};
this.xrCameraClipFromView ??= new THREE.Matrix4();
this.xrCameraWorldFromView ??= new THREE.Matrix4();
this.xrCameraClipFromView.fromArray(xrView.projectionMatrix);
this.xrCameraWorldFromView.fromArray(xrView.transform.matrix);
this.hasXRCameraParams = true;
```

then the two table lookups prefer those if they're set:

```ts
export function getDeviceCameraClipFromView(renderCamera, deviceCamera, targetDevice) {
  if (deviceCamera.simulatorCamera) { /* ...unchanged... */ }
  if (deviceCamera.hasXRCameraParams) {
    return deviceCamera.xrCameraClipFromView; // intrinsics from WebXR
  }
  return DEVICE_CAMERA_PARAMETERS[targetDevice].projectionMatrix; // fallback
}

export function getDeviceCameraWorldFromView(renderCamera, xrCameras, deviceCamera, targetDevice) {
  if (deviceCamera?.simulatorCamera) { return renderCamera.matrixWorld.clone(); }
  if (deviceCamera.hasXRCameraParams) {
    return deviceCamera.xrCameraWorldFromView.clone(); // pose from WebXR
  }
  // ...existing DEVICE_CAMERA_PARAMETERS[targetDevice].getCameraPose fallback...
}
```

stuff i'd want to check on a real device before trusting it:
- does each device actually hand the rgb cam back as its own `XRView` (so `projectionMatrix` is the camera's, not the display's), or reuse a render view. The spec says it should make a separate view if it's not aligned?
- is raw camera access granted on Aura / Galaxy XR for the center rgb cam?

for this PR i'd keep the table + UA as the stopgap (the Quest 3 numbers are what i measured/tuned), and do the XRView intrinsics thing as a follow-up with the table as fallback?

dli7319 (approved):

PR #269, opened 2026-05-10
Remove duplicate rollup build from serve script

nsalminen:
Resolves a compilation race condition introduced in 4d3b8fc7 where running npm start spawned two concurrent rollup instances attempting to write to build/xrblocks.js simultaneously.


dli7319 (approved):

ruofeidu:
Thanks Nels! I think we shall go through our current sdk next week for maintenance!

PR #273, opened 2026-05-11
Update @sparkjsdev/spark to v2.0.0 and optimize XR rendering

nsalminen:
- Upgrade @sparkjsdev/spark dependency and CDN import maps to v2.0.0.
- Reduce SparkRenderer maxStdDev from Math.sqrt(5) to Math.sqrt(4) to optimize WebXR performance.
- Update Simulator to toggle encodeLinear directly on SparkRenderer for v2.0 API compatibility.

Tested in simulator and Galaxy XR.

### DIFF
--- a/src/simulator/Simulator.ts
+++ b/src/simulator/Simulator.ts
@@ -75,7 +75,7 @@ export class Simulator extends Script {
   private initialized = false;
   private renderSimulatorSceneToCanvasBound =
     this.renderSimulatorSceneToCanvas.bind(this);
-  private sparkRenderer?: SparkRenderer & {encodeLinear?: boolean};
+  private sparkRenderer?: SparkRenderer;
   private registry?: Registry;
 
   constructor(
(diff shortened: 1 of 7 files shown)


src/simulator/Simulator.ts
```diff
@@ -75,7 +75,7 @@ export class Simulator extends Script {
   private initialized = false;
   private renderSimulatorSceneToCanvasBound =
     this.renderSimulatorSceneToCanvas.bind(this);
-  private sparkRenderer?: SparkRenderer;
+  private sparkRenderer?: SparkRenderer & {encodeLinear?: boolean};
```
dli7319: Is encodeLinear missing from the type?

dli7319 (approved):

src/simulator/Simulator.ts
nsalminen: Thanks for catching that! That's not necessary so removing.

dli7319 (approved):

PR #274, opened 2026-05-11
language detector demo - uses gemini live for the audio stream + mediapipe's language detector

salmanmkc:
new demo: spatial transcript that detects the language per utterance.

uses gemini live for the audio stream + mediapipe's language detector to tag each utterance. lays out as a chat-style list with the latest at the bottom — slots scale together so it doesn't jump when a new line comes in.

works for cjk / devanagari etc since the lineCount comes from troika after the text sync, so 2-line wraps don't overflow into the next slot.

all under demos/language_detector/, no engine changes.

Live demo for 10 minutes: https://youtu.be/hr1m-uBq6Os


ruofeidu (approved):
Amazing demo! Works like a charm!

ruofeidu:
Everything looks like a charm!

I forgot to mention that new demos would prefer https://xrblocks.github.io/samples/uiblocks/ for UI, but this one looks already very good

salmanmkc:
> Everything looks like a charm!
> 
> I forgot to mention that new demos would prefer https://xrblocks.github.io/samples/uiblocks/ for UI, but this one looks already very good

Awesome, glad you like it! 🥳  Will keep that in mind for future PRs!

salmanmkc:
This should be a sample?

PR #275, opened 2026-05-11
Make enableCamera also request microphone permission.

dli7319:


salmanmkc (approved):

salmanmkc:
ty for fixing this

nsalminen:
@dli7319 Is there a reason why we don't have a separate flag for enabling microphones instead? I intentionally left out microphone permissions here initially to prevent requesting unnecessary user permissions.

dli7319:
> @dli7319 Is there a reason why we don't have a separate flag for enabling microphones instead? I intentionally left out microphone permissions here initially to prevent requesting unnecessary user permissions.

Good point. Let me change it in the gemini icebreakers demo itself instead.
The prompt itself is actually different if we only request camera and not mic:
Screenshot 2026-05-11 at 4 06 12 PM
Screenshot 2026-05-11 at 4 07 56 PM

PR #277, opened 2026-05-11
Move microphone permission request from `enableCamera()` to gemini_icebreakers

dli7319:
This way the mic isn't requested if the app only needs the camera.

Reverts #275


nsalminen (approved):
LGTM! Should we also have a `options.enableMicrophone()`?

dli7319:
> LGTM! Should we also have a `options.enableMicrophone()`?

I don't think we need it. We don't have any special abstraction or wrapper for the microphone. This merely opens and closes the microphone briefly so the permission prompt shows up in Chrome.

PR #286, opened 2026-05-13
Attach onWindowResize listener earlier.

dli7319:
If the window resizes in the current async calls between creating the camera and attaching this listener, we might miss any window resize events.


ruofeidu:
Thanks David!

ruofeidu (approved):

PR #293, opened 2026-05-19
Catch exceptions in script lifecycle methods.

dli7319:
This catches all exceptions in script lifecycle methods and logs them as an error instead of crashing the update loop.

This is controlled by `options.catchScriptExceptions` which is now true by default.


ruofeidu (approved):
Awesome, thank you!

PR #291, opened 2026-05-14
Depth support for mobile.

dli7319:
1. Updated DepthOptions and Core to configure WebXR depth sensing using standard dataFormatPreference and usagePreference arrays.
2. Modified Depth, DepthTextures, and SimulatorDepth to dynamically read and pass depthDataFormat from the active XRSession instead of relying on static configuration.
3. Enhanced XREffects with dynamic camera viewport scaling to enable postprocessing pipelines on single-camera mobile AR devices.
4. Fixed missing luminance-alpha decoding across the depthmap sample and core shaders to properly unpack 16-bit integer depth on mobile devices.
5. Corrected mobile depth texture rotation and X/Y flipping in depthmap and DepthMesh by properly mapping WebGL UVs before applying WebXR's normDepthBufferFromNormView transform.

depthmesh, depthmap, and ballpit samples are now working on mobile.
vlcsnap-2026-05-14-10h49m17s569

PR #296, opened 2026-05-20
Fix depth mesh on Quest 3.

dli7319:
I accidentally broke this in #291.

Basically, Android on mobile only support luminance-alpha so I decided to pipe the depth format throughout the depth code so it supports both luminance alpha or float32 depending on what the session returns. However, when we convert GPU depth to CPU depth, the format gets converted to float32 so we need to hardcode that.

### DIFF
--- a/src/depth/Depth.ts
+++ b/src/depth/Depth.ts
@@ -298,14 +298,9 @@ export class Depth {
     }
   }
 
-  updateGPUDepthData(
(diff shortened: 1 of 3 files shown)


dli7319:
Note this assumes the gpu depth data will always be unsigned short on Quest so it might break again in the future if quest browser decides to support the requested gpu depth formats.

ruofeidu (approved):

PR #298, opened 2026-05-21
One dollar gesture estimator 2

samiriff:
Enable unistroke detection using the $1 unistroke detector.  It enables users to draw geometric shapes in the air using hand tracking and have them recognized in real-time. The PR also includes a comprehensive demo to showcase the feature, complete with visual tracing, shape completion, and animation effects.

To achieve this, a new SoundDetector class has been added, which uses a "Backend" strategy, allowing for easier expansion to other detection frameworks in the future.

- Modular Recognizer Backends: StrokeRecognition.ts delegates detection logic to specific backend classes (OneDollarUnistrokeRecognizer) using a StrokeRecognizerBackend class
- Configuration: Added StrokeRecognitionOptions.ts to tweak the recognition process.

Demo: https://youtu.be/gqTOH4F8rK4


ruofeidu:
Amazing! Thank you Salman!

David: would you also give a pass?

dli7319 (approved):

ruofeidu:
1. Currently it's projecting onto the camera's local XY plane, it will works well if user draws in front. (it could be another PR) For a more robust spatial experience, project the 3D points onto their own best-fitting 3D plane (calculated via singular value decomposition or a simple 3-point plane normal estimator) rather than the camera's viewport plane.

src/input/strokes/providers/OneDollarUnistrokeRecognizer.ts
```diff
@@ -0,0 +1,404 @@
...
+  const c = centroid(points);
+  const theta = Math.atan2(points[0].y - c.y, points[0].x - c.x);
+  return rotateBy(points, -theta);
+}
+
+/** Scales a list of points to a standard size (bounding box width/height becomes size). */
+function scaleTo(points: Point2D[], size: number): Point2D[] {
+  const B = boundingBox(points);
+  const newPoints = [];
+  for (let i = 0; i < points.length; i++) {
```
ruofeidu, line 138: // Prevents division by zero or extremely small dimensions.
```ts
  const EPSILON = 1e-5;
  const width = Math.max(B.width, EPSILON);
  const height = Math.max(B.height, EPSILON);
  for (let i = 0; i < points.length; i++) {
    const qx = points[i].x * (size / width);
    const qy = points[i].y * (size / height);
    newPoints.push({x: qx, y: qy});
  }
```

ruofeidu (approved):

src/input/strokes/providers/OneDollarUnistrokeRecognizer.ts
```diff
@@ -0,0 +1,404 @@
...
+    const pointsForwardRotated = this.preprocess(points, true);
+    const pointsBackwardRotated = this.preprocess(
+      points.slice().reverse(),
+      true
+    );
+    const pointsForwardUnrotated = this.preprocess(points, false);
+    const pointsBackwardUnrotated = this.preprocess(
+      points.slice().reverse(),
+      false
+    );
```
ruofeidu, line 244: Resampling, scaling, and translating only need to happen once on the forward path. The backward and rotated variants can be derived directly from that single preprocessed array:

override recognize(points: Point2D[]): StrokeRecognitionResult {
  // Preprocesses (resamples, scales, and translates to origin) the forward points once.
  const pointsForwardUnrotated = this.preprocess(points, false);
  
  // Reverses the preprocessed forward points to get the backward points.
  const pointsBackwardUnrotated = pointsForwardUnrotated.slice().reverse();
  // Rotates both forward and backward points to zero.
  const pointsForwardRotated = rotateToZero(pointsForwardUnrotated);
  const pointsBackwardRotated = rotateToZero(pointsBackwardUnrotated);
...

demos/unistroke_recognizer/PerfectShapeRenderer.js
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  /**
+   * Removes the shape from the scene and disposes of its geometries and materials.
+   */
+  dispose() {
+    if (!this.line) return;
+
+    this.scene.remove(this.line);
+    this.line.geometry.dispose();
```
ruofeidu: Dynamically allocating and disposing of GPU geometries causes frequent memory thrashing and WebGL buffer re-uploads, which can trigger micro-stutters.

Since shapes are static and predefined, cache and reuse the line and fill geometries:

```
export class PerfectShapeRenderer {
  // Caches geometries to avoid recreation and GPU upload overhead.
  static geometryCache = new Map();
  static getGeometries(name) {
    if (PerfectShapeRenderer.geometryCache.has(name)) {
      return PerfectShapeRenderer.geometryCache.get(name);
    }
    const shape = this.getShape(name);
    if (!shape) return null;
    // Generates line geometry.
    const points = name === 'Circle' ? shape.getPoints(32) : shape.getPoints();
    const lineGeom = new THREE.BufferGeometry().setFromPoints(points);
    // Generates fill geometry.
    if (name === 'V' || name === 'Caret') {
      shape.closePath();
    }
    const fillGeom = new THREE.ShapeGeometry(shape);
    const geometries = {lineGeom, fillGeom};
    PerfectShapeRenderer.geometryCache.set(name, geometries);
    return geometries;
  }
}
```

ruofeidu (approved):

src/input/strokes/providers/OneDollarUnistrokeRecognizer.ts
samiriff: Done.

demos/unistroke_recognizer/PerfectShapeRenderer.js
samiriff: Done.

src/input/strokes/providers/OneDollarUnistrokeRecognizer.ts
samiriff: Since scaling is non-uniform (c.f. https://faculty.washington.edu/wobbrock/pubs/uist-07.01.pdf), there might be issues in reordering these steps since the $1 algorithm resamples -> rotates -> scales -> translates. One option is to switch to uniform scaling. Is that fine?

In the meantime, I have refactored this code to reduce duplicate resampling and centroid calculations.

qxziuan (approved):
Cool demo! LGTM.

PR #304, opened 2026-05-26
Upgrade @sparkjsdev/spark to 2.1.0 to fix shader issue

nsalminen:
Fixes a GLSL shader compilation crash caused by the 'packed' reserved keyword in v2.0.0. Also updates importmap CDN URLs in walkthrough and modelviewer demos.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,10 +74,5 @@
   },
   "peerDependencies": {
     "three": "^0.184.0"
-  },
-  "overrides": {
-    "@sparkjsdev/spark": {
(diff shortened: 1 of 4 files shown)


dli7319 (approved):

package.json
dli7319: Can we remove this now?

ruofeidu (approved):

package.json
nsalminen: Good call! Will remove it. Thanks!

PR #305, opened 2026-05-26
Add dependency caching, test runs, and precise path triggers in workflows

nsalminen:
Optimize Github Action workflow execution efficiency by applying node_modules caching and precise path filters to the workflow checks. Also adds a testing step in the build workflow.

- Implement actions/cache@v4 to cache node_modules in build-check, lint-format-check, docs-build-check, and build-branch workflows.
- Add a `npm run test` step in the build-check workflow.
- Add path filters to build-check and lint-format-check to prevent unnecessary executions on unrelated assets.
- Constrain docs-build-check execution to documentation, core source, and configuration updates.

### DIFF
--- a/.github/workflows/build-branch.yml
+++ b/.github/workflows/build-branch.yml
@@ -10,6 +10,9 @@ on:
       - 'src/**'
       - 'package.json'
       - 'package-lock.json'
+      - 'tsconfig*.json'
+      - 'rollup.config.js'
+      - '.github/workflows/build-branch.yml'
 
 permissions:
   contents: write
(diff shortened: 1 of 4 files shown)


ruofeidu (approved):
Thank you Nels for optimizing the productivity!!

dli7319:
IIUC, `actions/setup-node@v6` is already supposed to do caching.
Is this not true?

nsalminen:
> IIUC, `actions/setup-node@v6` is already supposed to do caching. Is this not true?

@dli7319 As far as I know, `actions/setup-node` only caches global package tarballs to save download time, but still runs `npm ci` on every build. That unzipping, linking, and post-install execution adds about 45 seconds of runner overhead. The two methods are complementary: setup-node optimizes network downloads on cache misses, while our actions/cache setup restores node_modules instantly on cache hits, skipping the install step entirely.

dli7319 (approved):
Good find!

PR #309, opened 2026-05-27
Upgrade to TypeScript 6.0

dli7319:
This also gives us access to the Temporal API supported from Chrome 148:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#new-types-for-temporal


ruofeidu (approved):
Nice!

PR #310, opened 2026-05-28
Autoplace on horizontal surface API.

dli7319:
Adds a `xb.world.autoplaceOnHorizontalSurface` to place objects on horizontal planes. Updates the modelviewer sample to leverage this API to place each object. Replaces the cylinder in the model viewer sample with an animated torus to be more interesting.

- Support optional semantic `label` on simulator planes in `SimulatorPlane`, `DetectedPlane`, and `SimulatorWorld` to allow custom floor/table categorization in emulator scenes.
- Implement `getObjectBoundingBox` in `HorizontalAutoplacement` to exclude `ModelViewer` interaction handles (platform, rotation mesh) from collision boxes and bounding offsets.
- Reuse pre-calculated `bbox` in `getObjectBoundingBox` to cleanly support Gaussian Splatting models without disappearing offsets.
- Increase sampling `gridSteps` to 15 in `HorizontalAutoplacement` for tighter, high-precision object packing on tables.
- Register simulated plane meshes inside the `PlaneDetector` map to allow dynamic targeting and visualization.
- Swap cylinder mesh with animated, electric-teal `TorusKnotGeometry` in `ModelViewerScene` sample.

https://github.com/user-attachments/assets/52e5ccce-7e6e-4e57-85c9-1ea3e08aae59

### DIFF
--- a/samples/modelviewer/ModelViewerScene.js
+++ b/samples/modelviewer/ModelViewerScene.js
@@ -10,15 +10,16 @@ const PROPRIETARY_ASSETS_BASE_URL =
   'https://cdn.jsdelivr.net/gh/xrblocks/proprietary-assets@main/';
 
 export class ModelViewerScene extends xb.Script {
-  constructor() {
-    super();
-  }
+  loadedObjects = [];
+  placedObjects = new Set();
+  sessionStarted = false;
+  torusMesh = null;
 
   async init() {
     xb.core.input.addReticles();
     this.addLights();
     this.createModelFromObject();
-    await Promise.all([
+    return Promise.all([
       this.createModelFromGLTF(),
       this.createModelFromAnimatedGLTF(),
       this.createModelFromSplat(),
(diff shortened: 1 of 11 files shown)


samples/modelviewer/main.js
```diff
@@ -18,5 +18,7 @@ document.addEventListener('DOMContentLoaded', async () => {
     },
   ];
   options.setAppTitle('Model Viewer');
+  options.world.enabled = true;
```
dli7319: Btw, I think this is unnecessary. Can I remove it?

dli7319: I mean having `.enabled` on WorldOptions.

samples/modelviewer/ModelViewerScene.js
```diff
@@ -33,19 +34,59 @@ export class ModelViewerScene extends xb.Script {
...
+      this.autoplaceObject(model);
+    }
+  }
+
+  async autoplaceObject(model) {
+    if (!this.sessionStarted) return;
+
+    if (this.placedObjects.has(model)) return;
+    this.placedObjects.add(model);
+    const placed = await xb.world.autoplaceOnHorizontalSurface(model, {
```
ruofeidu: auto sounds redundant here.
it could be `placeOnBestSurface` ?

src/world/World.ts
```diff
@@ -198,6 +217,42 @@ export class World extends Script {
...
+   * Automatically places an object onto a suitable horizontal plane in the environment.
+   * It prioritizes planes in front of the user, prefers tables/elevated surfaces over floors,
+   * and ensures the object does not intersect other existing objects or other planes in the scene.
+   * If placement fails in the current frame, it continues retrying frame-by-frame until the timeout is reached.
+   *
+   * @param objectToPlace - The Three.js Object3D to place.
+   * @param timeout - Optional timeout duration as a Temporal.Duration or Temporal.DurationLike object (defaults to 500ms).
+   * @returns A promise resolving to true if successfully placed, false otherwise.
+   */
+  async autoplaceOnHorizontalSurface(
```
ruofeidu: placeOnBestSurface?

samples/modelviewer/main.js
ruofeidu: Sure!

I don't think world should be explicitly enabled here --- this is a basic example without virtual / real world interaction.

dli7319: The placement requires plane tracking.

But I mean `world` itself doesn't do anything unless you also enable plane tracking or scene meshing so it seems redundant.

samples/modelviewer/ModelViewerScene.js
dli7319: I'll remove `auto`.
I think it's better to keep horizontal in case we want to have `placeOnVerticalSurface` in the future.

PR #312, opened 2026-05-28
Remove path triggers for PRs.

dli7319:
I have branch protection rules set to requires these to pass before merging PRs. However, they get skipped if the PR doesn't trigger each rule and the PR gets stuck.

### DIFF
--- a/.github/workflows/build-check.yml
+++ b/.github/workflows/build-check.yml
@@ -15,16 +15,6 @@ on:
       - 'vitest.config.ts'
       - '.github/workflows/build-check.yml'
   pull_request:
-    paths:
-      - 'src/**'
-      - 'demos/**/*.ts'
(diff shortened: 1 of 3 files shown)


dli7319:
E.g. see https://github.com/google/xrblocks/pull/311
Screenshot 2026-05-28 at 10 28 31 AM

ruofeidu:
sounds good! also I wonder shall we remove the unused branches? I usually want to keep only 2-3 branches for simplicity...

ruofeidu (approved):

dli7319:
> sounds good! also I wonder shall we remove the unused branches? I usually want to keep only 2-3 branches for simplicity...

They need to be cleaned up after merging.
I typically make branches on my own fork now to avoid accidentally leaving a bunch of random branches on google/xrblocks.

PR #317, opened 2026-05-29
Add uikit options for automatic @pmndrs/uikit renderer setup

dli7319:
Introduces the UIKitOptions class that automatically configures Three.js's WebGLRenderer when @pmndrs/uikit is enabled (setting localClippingEnabled to true and registering reversePainterSortStable as the custom transparent sort). This also sets the stage to make uikit available to internal XR Blocks components in the future.

Before:
```javascript
import {reversePainterSortStable} from '@pmndrs/uikit';
// ...
await xb.init(options);
const renderer = xb.core.renderer;
renderer.localClippingEnabled = true;
renderer.setTransparentSort(reversePainterSortStable);
```

After:
```javascript
import * as uikit from '@pmndrs/uikit';
// ...
options.uikit.enable(uikit);
await xb.init(options);
```

### DIFF
--- a/samples/glasses_ui/main.ts
+++ b/samples/glasses_ui/main.ts
@@ -1,6 +1,6 @@
 import 'xrblocks/addons/simulator/SimulatorAddons.js';
 
-import {reversePainterSortStable} from '@pmndrs/uikit';
+import * as uikit from '@pmndrs/uikit';
 import * as THREE from 'three';
 import * as xb from 'xrblocks';
 import {SystemUI} from 'xrblocks/addons/glasses/ui/SystemUI.js';
@@ -86,10 +86,7 @@ document.addEventListener('DOMContentLoaded', async () => {
   options.simulator.instructions.enabled = false;
   options.simulator.handPosePanel.enabled = false;
   options.simulator.renderToRenderTexture = false;
+  options.uikit.enable(uikit);
   xb.add(new GlassesUISample());
   await xb.init(options);
-  // Setup for @pmndrs/uikit.
-  const renderer = xb.core.renderer;
-  renderer.localClippingEnabled = true;
-  renderer.setTransparentSort(reversePainterSortStable);
 });
--- a/samples/uiblocks/index.html
+++ b/samples/uiblocks/index.html
@@ -33,7 +33,7 @@
 
   <body>
     <script type="module">
-      import {reversePainterSortStable} from '@pmndrs/uikit';
+      import * as uikit from '@pmndrs/uikit';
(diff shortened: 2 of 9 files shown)


samples/glasses_ui/main.ts
```diff
@@ -1,6 +1,6 @@
 import 'xrblocks/addons/simulator/SimulatorAddons.js';
 
-import {reversePainterSortStable} from '@pmndrs/uikit';
+import * as uikit from '@pmndrs/uikit';
```
ruofeidu, line 3: @qxziuan please discuss with @dli7319  regarding this change

samples/glasses_ui/main.ts
```diff
@@ -81,10 +81,7 @@ document.addEventListener('DOMContentLoaded', async () => {
   options.simulator.instructions.enabled = false;
   options.simulator.handPosePanel.enabled = false;
   options.simulator.renderToRenderTexture = false;
+  options.uikit.enable(uikit);
```
qxziuan, line 89: My thoughts, open to discussion.

If we already know we depend on uikit, do we need this option?

Or logically, if only when we want to use uiblocks do we need uikit, then can we set something like options.ui.backend = ‘uiblocks’ (or default) and we set uikit dependencies accordingly. I feel the xrblocks options doesn’t need to have knowledge about uikit, but just what ui it uses. Then the ui manages the dependencies.

In the future when we fully move to uiblocks we can then delete this flag.

dli7319: I'm trying to move these two lines from the app into XR Blocks.
```
renderer.localClippingEnabled = true;
renderer.setTransparentSort(reversePainterSortStable);
```

XR Blocks doesn't have a hard dependency on UI Blocks so we don't already know if it's available ahead of time. Alternatively, XR Blocks could try to detect it with a dynamic import. But this leaves the option out of the user's control.

qxziuan: Got it.

ruofeidu (approved):

samples/glasses_ui/main.ts
ruofeidu: I'm also ok with dynamic loading, or recommend user to enable uikit for complex ui?

dli7319: If it's ok with you, I prefer this method since it doesn't add a bunch of boilerplate into XR Blocks itself.

ruofeidu: sgtm

PR #318, opened 2026-05-29
Add emoji support to Glasses UI.

dli7319:
Created a `TextWithEmoji` uikit component which splits up text word-by-word and renders emojis with SVGs. This is a drop-in replacement for uikit's `Text` component but has some layout limitations and performance drawbacks which is why I'm only using it the in glasses ui.

image


qxziuan (approved):
LGTM

PR #321, opened 2026-05-30
Forward Kinematics on Simulation Hands for AI Gen

whuang37:
Noticed in SDK that the current simulation hands implementation takes hard coded translation + quaternion representation of hand pose joints which isn't as human readable and AI authorable. Should enable better interaction test cases (i.e. defining interaction traces to evaluate an XR scene) and AI interactions within XR blocks later on. Would also enable better LERP across gestures in the simulator.

Added the following to the SDK: 
- ```src/simulator/handPoses/NeutralHandPose.ts``` as a new base pose and forward kinematics (FK) base pose for rotation driven hand authoring. Just a flatter hand vs relaxed.
- Forward kinematics (FK) resolution in ```src/simulator/handPosesHandPoseFK.ts```
	- Using WebXR joint order, builds a set of local offset/rotations from a neutral hand pose. 
	- Reformats from translation to a set of euler angles (x,y,z) which define flexion/extension, abduction/adduction, and twist for each joint
	- Automatically convert by handedness so same rotation json can drive either hand
- Simulator apis to run rotations ```simulator.hands.setLeftHandRotations(rotations)``` and ```simulator.hands.setRightHandRotations(rotations)```
- ```parseSimulatorHandPoseRotations(json)``` to normalize AI/authored JSONS into safe simulator hand rotations.

Created the demo ```demos/sim_hand_poses``` which provides a default simulator scene with
- sliders to control rotation of every joint for each hand
- heuristic gesture detector implemented to check compatability with existing gesture detection
- Gemini gesture generation authoring to demo capabilities of this new rotation format

Demo video: [https://youtu.be/onwPgNaAaC4](https://youtu.be/onwPgNaAaC4)
Thumb MCP can be a bit finnicky right now cause of orientation so its been muted in the system prompt of the demo. 

Noticed some smaller issues where the left and right hand actually have slightly different default pose definitions in relaxed which sometimes can cause issues. Full integration would rewrite all the other default gestures as rotation matrices on top of the neutral pose but since L/R hands aren't 1:1 mirrored copies of each other didn't implement this for now.


ruofeidu (approved):
Really nice commits. A README.md can be added with a brief intro in the demo folder.

ruofeidu:
@zhongyi-zhou  @qxziuan  take a look as well?

dli7319 (approved):
Awesome!

PR #323, opened 2026-06-01
Sim Hand Pose Migration to Angular Representation

whuang37:
Addressing some of the issues identified from the migration to FK angles in sim hands. 

Updated the following in the SDK:
- Established concrete sign conventions for hand flexion, extension, etc. following biomechanical principles
- Fixed the neutral hand pose such that both L/R hands are the same just mirrored
- Added a flag to establish biomechanical constraint ranges onto each hand angle
- Migrated all the preset poses to angular representations and made LERP across angles

Updated ```demos/sim_hand_poses``` to:
- follow the angles of the current preset pose

The migration to angles makes a small but nice change to the visualizations where now changing poses in the sim looks way more realistic since we LERP over angles. You can see this in the demo video below. First 9 seconds are the old representation rest of the video is the new representation.

[https://youtu.be/ihSz8tLXd1U](https://youtu.be/ihSz8tLXd1U)


demos/sim_hand_poses/main.js
```diff
@@ -519,27 +548,37 @@ async function start() {
...
+      rotations: cleanRotationsForJson(getActiveHandRotations()),
     })
   );
 
+  const syncDisplayedPose = () => {
+    syncControlsToRotations();
+    updateJsonViews();
+    requestAnimationFrame(syncDisplayedPose);
+  };
+  syncDisplayedPose();
```
dli7319: Can we call this in `update()` instead?

dli7319 (approved):

demos/sim_hand_poses/main.js
whuang37: just moved this out into its own xb script + update

PR #330, opened 2026-06-02
fix(uiblocks): prevent leading spaces on wrapped lines in TextWithEmoji

dli7319:
- Automatically groups space segments with their preceding words by appending spaces directly to the word's Text value.
- Groups space segments with their preceding emojis by applying calculated spaces as marginRight on the emoji's Image component.
- Retains explicit spaces following newlines or at the start of text as standalone space segments.
- Updates and expands Vitest suite with comprehensive space-grouping layout assertions.

image


qxziuan (approved):
LGTM.

PR #338, opened 2026-06-04
Add lipsync addon: audio-driven mouth animation for avatars

salmanmkc:
Recently I built a standalone audio avatar lipsync demo: https://salmanmkc.github.io/audio-avatar-lipsync/, open it on a phone/laptop, talk into the mic, watch the mouth move. This PR brings the same idea into xrblocks as a proper `lipsync` addon, with the integration I actually care about: every netblocks peer's voice driving their own avatar's mouth in the shared room.

Prelim demo: I will record a better one probably

https://github.com/user-attachments/assets/2be70908-5117-4fd4-ad3e-80c91ffe92df

The hard part isn't the rendering, it's getting "voice goes in, mouth moves in time" to feel responsive without shipping a model. So this PR does it with pure DSP: `MediaStream` feeds an `AnalyserNode`, then MFCC + F1/F2 formants + RMS/voicing per frame, then heuristic mapping to six viseme weights (jawOpen, aa, oo, oh, ee, consonant) with frame-rate-independent smoothing so 72/90/120 Hz Quest looks identical to 60 Hz desktop. The face itself is rendered as a flat canvas decal anchored to the front of the head sphere, a single dark ellipse that morphs, plus two eye dots and an occasional blink so a bare avatar reads as a face.

Where the face lives: `StylizedFace` is a small visual primitive that lives in xrblocks core alongside the `VisemeWeights` contract type. Netblocks `RemoteUserAvatar` now ships with a `face: StylizedFace` by default, parented to the default head so `defaultMesh.visible = false` cleanly hides it for custom avatars. Lipsync stays focused on audio. `LipsyncMouth` is a `Script` that reads a stream, computes visemes, and writes them to any target with `setVisemes(VisemeWeights)`. The lipsync addon does not own a visual of its own; the netblocks avatar's face is the target in multiplayer, and the puppet sample builds a `StylizedFace` directly.

Minimal usage (standalone puppet):

```ts
import * as xb from 'xrblocks';
import {LipsyncMouth} from 'xrblocks/addons/lipsync';

const face = new xb.StylizedFace({showEyes: false});
puppetHead.add(face);
const driver = new LipsyncMouth(micStream, {target: face});
puppetHead.add(driver);
```

For netblocks (the actual reason this exists; avatars already have a face, so just drive it):

```ts
session.voice.onTrack((peerId, stream) => {
  const user = session.users.get(peerId);
  if (!user) return;
  const driver = new LipsyncMouth(stream, {
    target: user.avatar.face,
    audioContext: THREE.AudioContext.getContext(),
  });
  user.avatar.add(driver);
});
```

The shared `audioContext` matters because browsers cap contexts around six per page; one per peer doesn't scale.

### Netblocks improvements that came with this

This PR also tightens a few netblocks edges that the lipsync sample surfaced:

- WebRTC voice now uses explicit `bye` and `hello` signals so `disable()` cleanly tears down the remote PC and a subsequent `enable()` re-negotiates. Previously orphan PCs collided with new offers and audio silently failed even though both UIs showed "voice on".
- `NetSession` dispatches `local-voice-state` and broadcasts `netblocks/voice-state` so any sample can render a reliable mic indicator (`RemoteUserAvatar` appends a 🎙️ on the name label) without depending on per-browser WebRTC track event timing.
- A `pagehide` listener now sends a session bye when a tab closes so remote peers see the avatar disappear immediately instead of waiting 15 to 30 seconds for WebRTC ICE to time out and notice.
- `enable()` snapshots a generation counter before awaiting `getUserMedia` so a fast disable mid-permission cancels the in-flight mic acquisition.

### Samples

- `src/addons/lipsync/samples/puppet/`: single user, mic into a puppet head. No server needed; opens in ten seconds.
- `src/addons/lipsync/samples/netblocks/`: multiplayer, each peer's voice drives the face on their netblocks avatar. Open in two tabs (or two devices) to see it.
- All seven existing netblocks samples now show faces on every peer by default, no sample wiring needed (free from the `RemoteUserAvatar` change).
- The netblocks samples index has a new "See also" link to the lipsync multiplayer demo.

### Scope and follow-ups

This PR is heuristic-only on the audio side. The standalone demo above also trains a tiny in-browser model from webcam + mic and uses that at inference time; the architecture here is set up so an `MLVisemeMapper` could slot into the same `AudioFeatures` to `VisemeWeights` interface as `FormantVisemeMapper` in a follow-up PR. Happy to keep the heuristic as the default (no download, runs everywhere) and add the model path as an opt-in.

### Lifecycle and ownership

The caller owns the `MediaStream`, any injected `AudioContext`, and the `target` face. `LipsyncMouth.dispose()` disconnects audio nodes, resets the target to rest (so a peer who stops speaking mid-vowel never leaves their face frozen open), and releases internal buffers. It does not stop tracks, close caller-supplied contexts, or dispose the target. `StylizedFace.dispose()` is idempotent so xrblocks' scripts manager and a host can both safely call it.

### Tested

- 253/253 unit tests pass; new coverage spans BlendshapeReducer, MfccExtractor, computeAudioFeatures, FormantVisemeMapper, StylizedFace, LipsyncMouth, VoiceChat (bye/hello, onLocalStateChange, enable race), RemoteUserAvatar (face attachment, custom-avatar opt-out, dispose), and NetSession (close ordering, pagehide).
- `npm run lint`, `npm run format:check`, `npm run build` clean.
- Verified locally cross-device (desktop Chrome with two tabs, and Android Chrome).
- Quest 3 device test still pending, and android XR, I don't have one on me at the moment.
- Demo video coming soon.


ruofeidu:
I'm impressed by this demo, thanks Salman!

salmanmkc:
> I'm impressed by this demo, thanks Salman!

my pleasure! glad you like it!

dli7319:
Super cool!
Should go well with @Peter-Pater 's VRM demo if we can integrate it with Gemini Live.

src/StylizedFace.ts
```diff
@@ -0,0 +1,244 @@
...
+      this.blinkStartAt = now;
+      // Random interval between blinks: 2.5–6.5 seconds, in the
+      // ballpark of natural human blink rate (15–20 per minute).
+      this.nextBlinkAt = now + 2500 + Math.random() * 4000;
+    }
+    return 1;
+  }
+}
+
+function clamp(x: number, lo: number, hi: number): number {
```
dli7319: Fyi we have a clamp already in `src/utils/utils.ts`
Or you can use threejs clamp: https://threejs.org/docs/#MathUtils.clamp

dli7319 (approved):

src/StylizedFace.ts
salmanmkc: Good call, thanks. Switched all three sites (StylizedFace.ts, BlendshapeReducer.ts, FormantVisemeMapper.ts) to the core clamp from src/utils/utils.ts.

dli7319 (approved):

src/StylizedFace.ts
salmanmkc: Actually scoped it down to just `StylizedFace.ts` in d54b317.

Tried it in the other two as well, but `BlendshapeReducer.ts` and `FormantVisemeMapper.ts` are addons, so they can only reach core via the `'xrblocks'` barrel. The barrel runs `new Core()` (and a real `AudioContext`) as a side effect at import time, which means the swap pulls the full core runtime in every time those DSP files load, plus jsdom can't construct an AudioContext so the tests fall over. Felt like more weight than the readability win was worth. Happy to revisit if you'd prefer consistency anyway.

dli7319: This is good, thanks. I may look into it later if I decide that it bothers me.

ruofeidu:
We current have samples for lipsync and uiblocks under src, 
will it be compiled to addons? shall we move it to samples folder?

@salmanmkc  @dli7319

salmanmkc:
> We current have samples for lipsync and uiblocks under src, will it be compiled to addons? shall we move it to samples folder?
> 
> @salmanmkc @dli7319

good point thank you, addressed in #342

Peter-Pater:
> Super cool! Should go well with @Peter-Pater 's VRM demo if we can integrate it with Gemini Live.

Wow this is impressive. I can give it a try integrating it into the vrm demo.

PR #339, opened 2026-06-05
skills: Added top-level design for GEMINI, CONTEXT, AGENTS, SKILLS.

ruofeidu:
- Added top-level GEMINI.md, CONTEXT.md, AGENTS.md, SKILLS.md
- Added guidance to `skills/xb-ui/SKILL.md` to use `xb-uiblocks` when authoring advanced spatial UI.
- renamed skills.md to SKILL.md

I am following best practices from https://github.com/googleworkspace/cli,
with core design principles from our two white papers and three AI agents self-reflecting.

Human review-revise for two hours but may still have missing points.
Let's improve together to make our SDK modernly AI-native :)


ruofeidu:
We are missing glasses skills and better camera / stereo skills (yet the template are enough I think).

zhongyi-zhou:
Have we evaluated these skills with some benchmarks? How much performance improvement there is if LLM uses this skill? 

I would recommend doing some evals even on some small datasets before releasing the skills

ruofeidu:
This is a great idea!
Yet we don't enough people working on Coding Agent Evaluation --- do you have time to help?

dli7319 (approved):
These all seem quite short to me. We'll probably want to expand upon them in the future.

src/addons/uiblocks/SKILL.md
```diff
@@ -0,0 +1,95 @@
+---
+name: netblocks
```
dli7319: This file seems misplaced.
And the uiblocks skill got deleted.

dli7319 (changes requested):

src/addons/uiblocks/SKILL.md
ruofeidu: Thank you! Accidentally removed when git move somehow...!

dli7319 (approved):

skills/xb-depth/SKILL.md
```diff
@@ -0,0 +1,62 @@
...
+
+Depth gives virtual objects awareness of real-world geometry: occlusion, depth-aware
+reticles, and an optional depth-mesh collider for physics. See `samples/depthmap`,
+`samples/depthmesh`, and `demos/occlusion`.
+
+## Enable depth (with occlusion)
+
+```js
+const options = new xb.Options();
+options.enableDepth(); // depth sensing + depth mesh; virtual content is occluded by the world
```
salmanmkc: Small accuracy nit: the inline comment here says virtual content is occluded by the world, but `enableDepth()` only applies `xrDepthMeshOptions`, which doesn't touch `occlusion.enabled` ([`DepthOptions.ts:38`](https://github.com/google/xrblocks/blob/8d0e1b48540455d375837aedfd691c1500dd65ef/src/depth/DepthOptions.ts#L38) defaults it to `false`). The `demos/occlusion` sample sets `options.depth.occlusion.enabled = true` explicitly. wondering, should we have some comment to say to enable it or remove the default part?

ruofeidu: Great catch!

Yes, occlusion was disable now for generic use case since it may slow down FPS ;( 
Ideally the syntax should included occlusion but in reality you need to manually occlude things following the Occlusion sample.

CONTEXT.md
```diff
@@ -0,0 +1,103 @@
...
+options.enableXRTransitions(); // fade transitions
+options.enableVR(); // immersive-vr instead of immersive-ar
+```
+
+There is **no** `enablePhysics()` / `enableLighting()`:
+
+```js
+import RAPIER from '@dimforge/rapier3d-simd-compat';
+options.physics.RAPIER = RAPIER; // enables physics
+options.lighting = new xb.LightingOptions(xb.xrLightingOptions);
```
salmanmkc: don't know if missing something but `xb.xrLightingOptions` doesn't seem to be exported anywhere. `LightingOptions` is at https://github.com/google/xrblocks/blob/0de99fbfe43d2b8fc89bc330e4cc3090b3ed3f84/src/lighting/LightingOptions.ts but there's no preset of that name. `samples/lighting/main.js:17` uses it too, so the existing sample is in the same boat. Might be worth a separate tiny PR adding the preset to match `xrDepthMeshOptions`. thinking now the doc could use `new xb.LightingOptions()` with fields set inline if that's easier.

CLAUDE.md
```diff
@@ -0,0 +1 @@
+See [GEMINI.md](GEMINI.md).
```
salmanmkc, line 1: if Claude follows symlinks, this could just be a symlink to `GEMINI.md`. Not a big deal either way.

CONTEXT.md
dli7319: Good find! I'd prefer to just remove `xb.xrLightingOptions` from the skill and sample. The name doesn't sufficiently describe what this preset is supposed to be.

CLAUDE.md
ruofeidu: It does :)

skills/xb-ai/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+
+```js
+class Ask extends xb.Script {
+  async ask() {
+    if (!xb.ai.isAvailable()) return; // always guard
+    const res = await xb.ai.query({prompt: 'Write a haiku about dust.'});
+    console.log(res.text); // response text
+
+    // Multimodal: text + image parts
+    const res2 = await xb.ai.query({
```
salmanmkc, line 43: the public `AI.query` type is just `{prompt: string}` (https://github.com/google/xrblocks/blob/0de99fbfe43d2b8fc89bc330e4cc3090b3ed3f84/src/ai/AI.ts#L177-L180), so this multipart shape won't compile, even though `Gemini.ts` accepts it at runtime. want to widen the type, or trim this to a text-only example?

ruofeidu: I think it does? https://github.com/google/xrblocks/blob/main/demos/xrpoet/PoemGenerator.js

I wrote the very initial Gemini demo last year despite that things have change so much afterwards lol

AGENTS.md
```diff
@@ -0,0 +1,126 @@
...
+- Keep PR diffs focused and add/adjust colocated tests for new branches.
+
+## Adding a Feature / Subsystem
+
+1. Add config to the relevant `*Options` class (or a new one) and register it in `Core.init()`.
+2. Expose a chainable `enable*()` on `Options` if it is user-facing.
+3. Wire the subsystem into `Core` and the frame loop / DI `Registry`.
+4. Re-export public types from `src/xrblocks.ts`.
+5. Add colocated `*.test.ts`, a `samples/`/`templates/` example, and — for a distinct
+   capability — a `skills/xb-*` entry (see [`skills/README.md`](skills/README.md)).
```
salmanmkc, line 109: should we add others now since we're adding them in this pr?  `CONTEXT.md`, `src/SKILL.md`, addon README, and the `SKILL.md`

skills/xb-ai/SKILL.md
salmanmkc: yeah the runtime works fine, that's what `Gemini.ts` does. issue is just the public TS type on `AI.query`. the xrpoet demo is `.js` so it skips type checking, but a `.ts` consumer would get a compile error on `type: 'multiPart'` i would think, happy to send a tiny PR for it if useful.

CONTEXT.md
salmanmkc: sounds good, thanks for the quick pr fix!

skills/xb-ai/SKILL.md
ruofeidu: Thank you Salman for fixing that!

AGENTS.md
ruofeidu: Fixed. Thanks!

salmanmkc (approved):
great stuff! thanks for addressing all the comments!

salmanmkc:
ah actually still have xb.xrLightingOptions in the context.md file

ruofeidu:
> ah actually still have xb.xrLightingOptions in the context.md file

Fixed now! Thanks!

Lighting estimation works in Android XR, yet I'm not sure it is well supported in other platforms (likely not)
and the framerate on Android XR isn't as good as what it is on Android with ARCore.
Will add lighting context when we have more robust demos.

PR #343, opened 2026-06-05
widen AI.query input to accept GeminiQueryInput

salmanmkc:
per the thread on #338, the public input type on `AI.query` is just `{prompt: string}` so multipart / uri / base64 don't typecheck from a .ts caller, even though `Gemini.query` handles them fine at runtime.

widens it. also routes through the model so OpenAI throws a clear error instead of sending undefined content when given a non-prompt shape.


dli7319:
TY!

ruofeidu (approved):

PR #346, opened 2026-06-05
MaterialSymbolsView: read viewBox via getAttribute

salmanmkc:
`as any` was needed because `@types/three` mistypes `SVGResult.xml` as `XMLDocument` when at runtime it's actually `xml.documentElement` (Element).

dropped the cast, uses `getAttribute('viewBox')` instead of the quirky `.attributes.viewBox.value` named-property access. also `.map(Number)` on the split parts, since the old code relied on `as any` to launder the strings into numbers for the arithmetic below.


dli7319:
Good find!

PR #351, opened 2026-06-06
evals: harness for the xb-* skills (first cut)

salmanmkc:
first cut at a benchmark for the xrblocks `xb-*` skills, calls gemini via the api the way the canvas-gem deploys it. skill content baked into the system prompt, no filesystem access. the model has to produce a complete `main.js` from scratch.

20 tasks across the 13 `xb-*` skills, half phrased as engineer specs ("modify `main.js` to enable WebXR depth sensing"), half as the imaginative requests users actually type into the gem ("make my room feel like there's a giant statue standing across it"). same scoring rubric on both: import / api / parse / forbidden patterns + an llm-judge pass (gemini-2.5-pro again, full SKILL.md as ground truth, scoring `accomplishes_task`, `idiomatic_xrblocks`, `hallucination_severity`).

ran the full sweep on both `gemini-2.5-pro` and `gemini-2.5-flash`.

## headline: api_match

did the agent call the APIs the skill defines? this is the cleanest single-number signal, since parse_ok / forbidden_clean / import_match are mostly invariant (gemini writes valid js either way, etc).

| model | prompt style | with skill | without skill | Δ |
|---|---|---:|---:|---:|
| pro | engineer-spec | 1.00 | 0.07 | **+0.93** |
| pro | canvas-faithful | 1.00 | 0.03 | **+0.97** |
| flash | engineer-spec | 0.93 | 0.07 | +0.87 |
| flash | canvas-faithful | 0.97 | 0.03 | +0.93 |

api_match by model and prompt style

without the skill, both models call the right xrblocks APIs ~3-7% of the time. with the skill, 85-100%.

## composite score (multi-dimensional)

| model | prompt style | with skill | without skill | Δ |
|---|---|---:|---:|---:|
| pro | engineer-spec | 1.00 | 0.65 | +0.35 |
| pro | canvas-faithful | 1.00 | 0.65 | +0.36 |
| flash | engineer-spec | 0.98 | 0.65 | +0.33 |
| flash | canvas-faithful | 0.99 | 0.61 | **+0.38** |

skill effect by model and prompt style

unexpected: **canvas-faithful prompts widen the skill gap, not narrow it**. vague prompts force the model to lean on its system-prompt priors, which is exactly what the skill content is.

## hallucination breakdown

llm-judge classified each output as `none` / `minor` / `major` hallucination using the full SKILL.md as ground truth:

| model | with skill (none / minor / major) | without skill (none / minor / major) |
|---|:---:|:---:|
| pro | 15 / 1 / 4 | 3 / 0 / **17** |
| flash | 13 / 2 / 5 | 0 / 0 / **20** |

without the skill, flash produces a `major` hallucination on every task. consistent patterns: jsx-like `xr-scene`/`xr-room` for netblocks, fake `<xr-hands>` elements for hands, invented A-Frame `xrb-button` for ui, made-up `xr.ai.multimodal` for ai. several without-skill outputs imported from `xr-blocks@0.4.1` and similar packages that have never been published.

with the skill, most outputs are rated `none`. the remaining `major` cases are concentrated on `xb-modelviewer`, which has a documentation gap the eval surfaced.

llm judge: idiomatic xrblocks rating per task

composite per task across both models

per-metric breakdown:

pro metric grid

flash metric grid

## skill gaps surfaced

beyond the headline numbers, the eval flagged concrete improvements to chase:

1. **xb-modelviewer**: even with the skill loaded, gemini invents a `Loader` class that does not exist. 4-5 `major` hallucinations per model. the skill description mentions `loadGLTFModel` and `setupPlatform` but the quick-start skips the actual loader call, so the model fills in something plausible-looking. fix: add a minimal `ModelViewer` loader example to the skill.
2. **xb-netblocks + xb-lipsync import paths**: the skills used a directory import (`xrblocks/addons/<addon>/src`) that 404s on the `@build` cdn because browser importmaps do not have node-style index resolution. fixed in #349 (netblocks) and #350 (lipsync).
3. **xb-depth**: with-skill composite drops to 0.88 because the agent uses `enableDepth` but misses adjacent expected setup. skill could spell out the full recipe.
4. **xb-hands** (flash only): composite 0.83 for the same shape of issue. depth and hands skills could both use a "minimum viable setup" snippet at the top.

## what's in the PR

- `evals/prototypes/runners/run_gem_api.py` - agent runner (gemini api directly)
- `evals/prototypes/score_proto.py` - scorer (skips vacuous import_match when expected_imports is empty)
- `evals/prototypes/judge.py` - llm-as-judge
- `evals/prototypes/smoke.py` - playwright + headless chromium (caught real CDN-resolution bugs in the netblocks and lipsync skills; fixed in #349 and #350)
- `evals/prototypes/ablate.py` - drop one skill section at a time (binary scorer is overdetermined for this; would benefit from judge in the loop)
- `evals/run_all.sh` + `summarize_proto.py` + `plot.py` - orchestrator, summary, matplotlib charts
- `evals/prototypes/tasks/` - 20 tasks (10 engineer-spec + 10 canvas-faithful)

## what's not in the PR

- multi-trial cells (each cell ran once, no IQR)
- cross-model judge (judge is gemini-pro, same family as the agent)
- runtime correctness beyond load-time smoke
- multi-skill composition tasks
- skill-edit regression CI

first attempt, posting for design feedback before scaling further.


ruofeidu:
I am deeply impressed!

As said in https://arxiv.org/abs/2603.24591, we also have an evaluation pipeline but haven't been open sourced since I don't think it it rigorous enough - also it was designed for traditional one-shot prompting instead of agentic loops. 

Thanks for the contribution, I'll take a deeper look and think of how we should open source of what we have with what you contribute here! I appreciate your proactivity and treat of as one of our key contributors here :)

salmanmkc:
> I am deeply impressed!
> 
> As said in https://arxiv.org/abs/2603.24591, we also have an evaluation pipeline but haven't been open sourced since I don't think it it rigorous enough - also it was designed for traditional one-shot prompting instead of agentic loops.
> 
> Thanks for the contribution, I'll take a deeper look and think of how we should open source of what we have with what you contribute here! I appreciate your proactivity and treat of as one of our key contributors here :)

My pleasure! Thanks Ruofei!I would have loved to run more tasks, however I'm using my own key for gemini. I was doing a more researchy style writeup on this but thought might be overkill for the PR.

dli7319:
Great work! A few thoughts from my side:
1. It seems pretty expected to me that Gemini base doesn't call the correct APIs automatically when no skills or system prompt is provided. The prompt as https://xrblocks.github.io/prompts/ would be the baseline for a fuller eval.
2. I like the prompts. They check specific skills and could be good references across versions of XR Blocks.
3. The implementation right now provides the specific skill with each prompt. Another important thing to check is whether Gemini picks the right skill on its own based on the name and description when given `listSkills` and `getSkill` functions to call.
4. The hard part is create an agent rater which runs in XR. This is blocking self-improvement of the actual generation and of XR Blocks.

dli7319 (dismissed):

ruofeidu:
> create an agent rater which runs in XR. This is blocking self-improvement of the actual generation and of XR Blocks

I see evolution of evaluation in this:

1. **Per-prompt evaluation**: Pass@1 rate, subjective / VLM evaluation.
2. **Per-task evaluation**: Anthony Chen is working on this.
3. **Per-skill evaluation** (this): similar to LLMR and AgentAR paper regarding API calls' successful rate
4. **Per-workflow evaluation**: asynchronizely, recursively, and holistically evaluate and improve the XR creation process. Asynchronizely improve the SDK skills when needed. @dli7319 proposed above.
5. **Gen-eval on-the-fly**: synchronically, recursively, and fast-slow-adaptively evaluate and improve the XR creation process. Asynchronizely improve the SDK skills when needed.

evals/prototypes/runners/run_gem_api.py
```diff
@@ -0,0 +1,172 @@
...
+    if usage:
+        for k in ("prompt_token_count", "candidates_token_count", "total_token_count"):
+            v = getattr(usage, k, None)
+            if v is not None:
+                usage_dict[k] = v
+    (log_dir / "usage.json").write_text(json.dumps(usage_dict, indent=2))
+
+    # Score using the existing scorer.
+    scorer = EVALS / "prototypes" / "score_proto.py"
+    result_dir = EVALS / "results" / model_slug / mode
```
ruofeidu, line 162: Here the it has model_slug, but 
summarize_proto.py
 still hardcodes path lookups directly to evals/results/gem-api-with-skill, causing it to fail with no results found when run.

ruofeidu (commented):
Please Create the requirements.txt file and update the setup docs in README.md

ruofeidu:
In a second thought regarding storing eval pictures: 
can we organize the charts folder into https://github.com/xrblocks/evals
and organize by xrblocks version & date with readme?

otherwise, the xrblocks SDK repo will grow exponentially :)

ruofeidu:
Action items for Salman:

1. Please Create the requirements.txt file and update the setup docs in README.md
2. summarize_proto.py still hardcodes path lookups directly to evals/results/gem-api-with-skill, causing it to fail with no results found when run.
3. Organize charts in a different PR to https://github.com/xrblocks/evals to reduce repo size.

dli7319:
Fyi, I typically prefer using pdf instead of png for charts. That way they're vectorized instead of rasterized and the text is selectable if we ever embed them into a white paper. Should also be smaller in file size as well.
The only con is that they can't be directly embedded in Google docs & sheets.

ruofeidu:
Agreed, always use vectorized graphics / pdfs for charts & figures!

salmanmkc:
thanks @ruofeidu + @dli7319, super helpful comments.

pushed:

- `requirements.txt` + a `pip install` line in the readme, and dropped the old gemini-cli runner.
- fixed `summarize_proto.py` (was reading the pre-per-model paths so it printed "no results found"), plus a small path guard in `run_gem_api.py`.

follow-ups, separate PRs:

- saw xrblocks/evals is already up, will move the charts there as PDFs and link from here.
- xrblocks.github.io/prompts/ as a third arm so we get with-skill / with-baseline / nothing.
- `listSkills`/`getSkill` agentic runner alongside the system-prompt one.
- the in-XR agent rater (per-workflow / gen-eval-on-the-fly in @ruofeidu's framing) is the more interesting long-term thing too, agree that's the unlock for self-improvement. happy to take a stab at a v0 once the above arms are in.

ruofeidu:
Can we refrain from checking in all these .png charts into the repo for now?

Feel free to create a PR to https://github.com/xrblocks/evals

We'd like to keep the repo size as small as possible (avoid having a 1GB+ repo in the long run :)

salmanmkc:
> Collaborator

Sure!

salmanmkc:
follow-up PR for the charts opened at xrblocks/evals#1. dropped `evals/charts/` from this repo (b90c625) earlier; the new repo holds them as vectorized PDFs alongside the raw scorer + judge JSONs and a `manifest.json` per run. happy for layout / convention feedback over there before it becomes the precedent.

ruofeidu (approved):

ruofeidu (approved):

ruofeidu (approved):

PR #354, opened 2026-06-08
Simulator: Add pointer lock mode.

dli7319:
Note the pointer gets hidden on my macbook. Not sure why it keeps showing up in the video recording.

https://github.com/user-attachments/assets/12f47f47-e726-411a-9338-d895871f743b

### DIFF
--- a/samples/pointer_lock/index.html
+++ b/samples/pointer_lock/index.html
@@ -1,0 +1,40 @@
+<!doctype html>
+<html lang="en">
+  <head>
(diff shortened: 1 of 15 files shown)


dli7319:
For other samples, this can be enabled from the simulator settings:

Screenshot 2026-06-08 at 4 32 33 PM

ruofeidu:
SG, this mode is mostly for XR gaming simulation.

PR #359, opened 2026-06-11
feat: implement missing features in ui, added test coverage, and revise logs

ruofeidu:
1. implemented onHover for TextButtons
2. add key check for Gemini Canvas
3. add support of `npm run test:coverage`

also manually tested all AI cases.


src/ui/components/TextButton.ts
```diff
@@ -119,7 +119,27 @@ export class TextButton extends TextView {
     this.textObj!.raycast = () => {};
   }
 
-  // TODO: Implement onHoverOver() and onHoverOut().
+  /**
+   * Handles behavior when the cursor hovers over the button.
+   */
+  onHoverOver() {
+    if (!this.ux) return;
+    this.update();
```
dli7319: Why do we need to call this.update()?
Doesn't update get called automatically?

src/utils/utils.ts
```diff
@@ -21,6 +21,20 @@ export function print(...args: unknown[]) {
   console.log('*', ...args);
 }
 
+const warnOnceMessages = new Set<string>();
```
dli7319: Won't this permanently take up memory?
Seems like a bad idea to me.

src/core/Core.ts
```diff
@@ -139,6 +140,24 @@ export class Core {
   webXRSessionManager?: WebXRSessionManager;
   permissionsManager = new PermissionsManager();
 
+  /**
+   * The WebGL renderer, created during {@link Core.init}. Reading it before
+   * `init()` has run returns `undefined` and logs a one-time warning.
+   */
+  get renderer(): THREE.WebGLRenderer {
+    if (!this._renderer) {
+      warnOnce(
```
dli7319: I think it makes sense to warn each time for this. There could be multiple scripts trying to access renderer.

src/core/Core.ts
```diff
@@ -139,6 +140,24 @@ export class Core {
   webXRSessionManager?: WebXRSessionManager;
   permissionsManager = new PermissionsManager();
 
+  /**
+   * The WebGL renderer, created during {@link Core.init}. Reading it before
+   * `init()` has run returns `undefined` and logs a one-time warning.
+   */
+  get renderer(): THREE.WebGLRenderer {
```
dli7319: We should probably change the return to `THREE.WebGLRenderer|undefined`

src/ui/components/TextButton.ts
ruofeidu: Removed! I didn't remember why I had the comment lol

src/core/Core.ts
ruofeidu: hopefully people / AI only make one mistake --- usually this is called in update I don't want to pollute the console :)

src/utils/utils.ts
ruofeidu: Agreed!

ruofeidu:
I reverted the suggestion of adding `undefined` to get renderer since we have too many places assuming renderer is ready. I think adding a console warning is already useful in case of accidentally fetching renderer without init.

ruofeidu:
Thanks for the double checking!

dli7319 (approved):

PR #363, opened 2026-06-12
objects_3d: single-press 3D object detection demo

salmanmkc:
adds a single-press 3D object detection demo: gemini for open-vocab labels, slimsam-77 for tight masks, depth-mesh raycasts + PCA to fit oriented 3D bounding boxes per object. press detect, get labels + boxes for everything in view.

for now just a demo, but if you want I could turn this into an addon

a few things worth flagging:

- the snapshot is 512x512 in the sim but the camera is 16:9, so a bbox at u=0 doesn't actually correspond to camera NDC x=-1. uvToNdc contracts along the mismatched axis, otherwise every ray goes ~78% too wide outboard and every box floats next to its object instead of on it. real quest cameras shouldn't have this mismatch but the helper is no-op when they match.
- camera + depth-mesh are frozen at detect start. detect takes seconds, the live camera drifts (mouse-look in sim, head motion on quest), and rays + mesh state would otherwise be out of sync with the snapshot pixels.
- multi-view fusion: pressing again from a new angle merges into existing boxes instead of replacing them. category match + centroid within combined half-extents = merge. fixed an over-aggressive merge radius that was collapsing 3 distinct pendant lamps into 1.
- three-mesh-bvh on the cloned depth mesh, pulled via importmap so no package.json churn. ~1000 stride-sampled rays per object × ~15 objects went from ~3 s to ~30 ms.
- per-object loop is pipelined: mask decodes serialize on the GPU (ORT session is single-threaded), but the cpu raycast/fit for object K now runs while the GPU decodes mask K+1. gemini + SAM warmup also run in parallel.
- defaults to gemini + slimsam since that gave the best per-press recall. mediapipe + slimsam is a deterministic no-api-key fallback via the dropdown.
- desktop sim only for now, quest hw untested (i'm not home atm)

better shown in screenshots:
Front view
Side view


dli7319:
It's working well in the simulator.
Have you tried it on device?

On my Galaxy XR, I'm getting a bunch of stuttering and weird errors.

ruofeidu:
Ah, you may have to try 

chrome://flags

and enable WebGPU and WebXR interop? 

It's still not officially landed yet.

dli7319 (approved):

salmanmkc:
> It's working well in the simulator. Have you tried it on device?
> 
> On my Galaxy XR, I'm getting a bunch of stuttering and weird errors: Screenshot 2026-06-15 at 10 49 35 AM

left my quest at home unfortunately, have been away

salmanmkc:
> Ah, you may have to try
> 
> chrome://flags
> 
> and enable WebGPU and WebXR interop?
> 
> It's still not officially landed yet.

great thanks!

salmanmkc:
I'll put the SDK usage in a follow up for bvh and some panel changes

nsalminen:
This is great @salmanmkc! Looking forward to trying it out on Galaxy XR

salmanmkc:
> This is great @salmanmkc! Looking forward to trying it out on Galaxy XR

Thank you @nsalminen! Let me know how it goes 😃

PR #364, opened 2026-06-12
world: add FaceRecognizer + face_mirror demo

salmanmkc:
adds xb.world.faces using MediaPipe's FaceLandmarker. the SDK now has objects, pose, and faces under the same backend-strategy pattern.

Each detected face exposes the 478-point face mesh in world space, optional 52 ARKit-style blendshape weights, and optional 4x4 rigid head pose. The blendshapes are the input the lipsync addon has been missing: until now there was no in-tree feed for them, so any avatar expression mirroring required wiring MediaPipe by hand in the demo. with this, `xb.world.faces` is the official source.

ships with a `face_mirror` demo: per-frame wireframe of the face mesh contours (face oval, lips, eyes, eyebrows) projected via the depth mesh, plus a live HUD of 12 high-signal blendshapes.

Desktop testability: the SDK reads from `xb.core.deviceCamera`, which on a Quest is passthrough and on desktop is the rendered sim scene. Neither shows a human face by default. the demo includes a small webcam fallback wrapper: when not in a WebXR session, call getUserMedia and route ImageData through a canvas. real-headset path is untouched. one gotcha worth flagging here too: don't mirror the canvas before feeding it to FaceLandmarker, it swaps every Left/Right blendshape and landmark index because the model uses subject-pov ARKit names. preview thumbnail mirrors via CSS instead.

On Quest the passthrough cameras look outward so the demo labels the faces of other people in front of you rather than your own. that's actually the more interesting use case (external avatar mirroring for video calls in XR, expression-driven NPCs, person-anchored UI, reading emotions, etc.).

### DIFF
--- a/demos/face_mirror/FaceMirror.js
+++ b/demos/face_mirror/FaceMirror.js
@@ -1,0 +1,380 @@
+import * as THREE from 'three';
+import * as xb from 'xrblocks';
+
+// Indices for the canonical face mesh tesselation that ships with the
+// MediaPipe FaceLandmarker model. We render a subset of triangles as line
+// segments so the wireframe reads clearly without overwhelming the scene.
+// These are the standard MediaPipe FACEMESH_TESSELATION index pairs for the
+// face mesh perimeter + key feature contours (lips, eyes, eyebrows). Source:
+// https://github.com/google-ai-edge/mediapipe-samples (face_landmarker)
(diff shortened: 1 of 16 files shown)


dli7319:
Thank you!

This will also need to be moved to a webworker.
The rendering FPS is very slow even on my m1 pro macbook.

dli7319 (approved):

PR #367, opened 2026-06-15
perf(world/faces): BVH-accelerate per-landmark depth raycasts

salmanmkc:
even with FaceLandmarker inference moved to a worker (#366), the per-detection raycast loop still tanked desktop FPS. perf trace showed `getVertexPosition` + `intersectTriangle` + `_computeIntersections` eating ~9 of every 14 main-thread seconds.

`processFaceLandmarkerResult` raycasts each of the 478 landmarks against the depth-mesh snapshot to project to world coords. stock three.js raycaster is O(triangles) per ray, depth mesh runs in the thousands of triangles, so the per-detection raycast loop alone dominates the frame budget.

wires three-mesh-bvh prototype patches at FaceRecognizer module load (same trick objects_3d uses) and calls `computeBoundsTree()` on the cloned depth-mesh geometry in `getDepthMeshSnapshot`. meshes without a BVH fall back to the stock walker so the patch is safe to apply globally.

measured improvement in the same trace before/after: `getVertexPosition` 4284ms -> 88ms, `intersectTriangle` 3211ms -> 57ms (~98% reduction on each).

follow-up commit caches the cloned snapshot + BVH and reuses them while the source geometry's position attribute version is unchanged (three.js bumps it on `needsUpdate = true`). world transform still refreshes each call so the cached snapshot reflects the current depth mesh pose. for a static desktop sim this amortizes the BVH build across all detections instead of running every detection.

three-mesh-bvh added to package.json + face_mirror demo importmap. same version objects_3d already pins.

related: #366 (worker), #370 (simulator depth inflight guard). all independent file-wise.

### DIFF
--- a/demos/face_mirror/index.html
+++ b/demos/face_mirror/index.html
@@ -32,7 +32,8 @@
           "uiblocks": "../../build/addons/uiblocks/src/index.js",
           "xrblocks": "../../build/xrblocks.js",
           "xrblocks/addons/": "../../build/addons/",
-          "@mediapipe/tasks-vision": "https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision@latest/vision_bundle.mjs"
+          "@mediapipe/tasks-vision": "https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision@latest/vision_bundle.mjs",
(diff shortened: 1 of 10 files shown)


dli7319:
I don't think we want `three-mesh-bvh` as a hard dependency requirement.
Is it possible to detect `three-mesh-bvh` availability like we do for other dependencies like troika?

dli7319 (approved):

PR #370, opened 2026-06-15
perf(simulator): don't stack depth-readback requests in flight

salmanmkc:
`SimulatorDepth.update()` was firing one render-to-target + one async readback per frame on the main thread. the readback uses `WebGLRenderer.readRenderTargetPixelsAsync` which polls a WebGL fence via setTimeout, and the fence-polling chain showed up dominating desktop perf traces of every depth-consuming demo (face_mirror, objects_3d, human_pose_detector). on a recent face_mirror trace this stack accounted for ~5 of every 12 main-thread seconds.

adds an inflight guard: don't fire a new readback while the previous promise is still resolving. without it `simulatorUpdate` stacks one per frame and the fence-polling chains overlap. when the in-flight readback resolves, the next `simulatorUpdate` runs a fresh pass normally.

initially also tried a motion gate (skip the pass when the depth camera hadn't moved), but two issues made it not worth it: scene-driven motion (animated object in SimulatorScene) wouldn't refresh depth until the camera moved, and slow tweens got skipped repeatedly then caught up in a burst. dropped it in the follow-up commit. inflight guard alone is the only behavior change.

real-headset path is untouched, `SimulatorDepth` only runs in the desktop simulator addon.

4 tests cover the guard: first-frame pass, no-double-fire while in flight, resume after readback resolves, steady-state per-frame.

context: surfaced while tracing face_mirror FPS for #366 / #367. independent of those, applies to every demo that consumes depth.

### DIFF
--- a/src/simulator/SimulatorDepth.test.ts
+++ b/src/simulator/SimulatorDepth.test.ts
@@ -3,18 +3,12 @@ import {beforeEach, describe, expect, it, vi} from 'vitest';
 
 import {SimulatorDepth} from './SimulatorDepth';
 
-// SimulatorDepth's update() does three things:
-//   1. Skip when an earlier async readback hasn't resolved yet
-//      (`updateInFlight`).
-//   2. Skip when the depth camera hasn't moved beyond the configured
-//      epsilons (`depthHasChanged`).
-//   3. Render the depth scene + start the async readback otherwise.
-//
-// These tests focus on the throttle logic. The actual WebGL render and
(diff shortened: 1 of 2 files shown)


dli7319 (approved):

dli7319:
Oh didn't know you're still working on this. Can you push your change again?

salmanmkc:
> Oh didn't know you're still working on this. Can you push your change again?

sure done sorry was just dropping the motion detection part, since it's purely camera based and scene could change itself instead, your commit state I think was fine but pushed anyway

PR #379, opened 2026-06-16
objects_3d: make the control panel draggable

salmanmkc:
the objects_3d control panel was head-leashed, so it followed your gaze around and you couldn't set it down anywhere. swapped HeadLeashBehavior for a draggable ManipulationBehavior so you can grab it and place it.

doing that surfaced two uiblocks rendering issues. the importmap was still pinned to `@pmndrs/uikit` 1.0.56 while the uiblocks addon is built against 1.0.64, so the button icons and labels didn't render at all. bumped the three `@pmndrs/*` pins to 1.0.64.

with that fixed the labels showed up but the deeply-nested button text was hidden behind its own backplate (the shared glyph mesh sits at the card plane while each button fill stacks a little forward, so the fill draws over the label), and the transparent fills sort by camera distance with depthWrite off so the button backgrounds blended inconsistently and flickered light/dark with the viewing angle. pinned the labels on top (depthTest off + higher renderOrder) and gave the button fills and icons a deterministic renderOrder, so it reads right from every angle now.

stacks on #363, rebase once that lands.

### DIFF
--- a/demos/objects_3d/index.html
+++ b/demos/objects_3d/index.html
@@ -27,9 +27,9 @@
           "webgl-sdf-generator": "https://esm.sh/webgl-sdf-generator@1.1.1/es2022/webgl-sdf-generator.mjs",
           "lit": "https://cdn.jsdelivr.net/gh/lit/dist@3/core/lit-core.min.js",
           "lit/": "https://esm.run/lit@3/",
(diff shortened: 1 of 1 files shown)


ruofeidu (approved):

ruofeidu (approved):

ruofeidu:
Thank you Salman!

salmanmkc:
> Thank you Salman!

you're welcome Ruofei!

PR #365, opened 2026-06-14
refactor: simplify FaceMirror orchestrator and modularize face landmarker demo components

dli7319:
Minor refactoring of @salmanmkc 's new face recognizer demo added in #364 .
No visible changes to the demo itself.

PR #386, opened 2026-06-17
fix face_mirror: camera-not-ready spam + stale depth-mesh cache

salmanmkc:
two separate things were breaking face_mirror in the simulator, both predating #365 (matches the report that the demo was broken even before that refactor). split into two commits.

first, the camera detectors (faces / humans / objects) run detection from frame 1, but the device-camera pose isn't ready until the simulator registers its camera or a WebXR session exposes its cameras. `getCameraParametersSnapshot` threw `No XR cameras available` during that window, so the console filled with ~10 errors on every startup. it now returns null while no pose is available and the detectors skip the cycle, same as they already do when the depth mesh isn't ready. a one-time warning still fires if the camera never shows up past a grace period so a real misconfig isn't silently swallowed.

second, the face wireframe only rendered while you moved the camera. `DepthMesh.update` writes fresh positions into the downsampled geometry but only bumped `needsUpdate` on the full-res geometry, so the downsampled geometry's position version never changed. `FaceRecognizer` caches a cloned depth mesh + BVH keyed on that version (added in f468793), so the cache never invalidated: the clone built over the initial empty geometry got reused forever, every per-landmark raycast missed, and all 478 landmarks collapsed onto the camera origin (so the wireframe sat on the camera, invisible). moving recreated the geometry object and forced a one-off rebuild, which is why it only flickered in on motion. one line: bump the downsampled geometry's `needsUpdate` too so the cache invalidates.

### DIFF
--- a/src/camera/CameraUtils.test.ts
+++ b/src/camera/CameraUtils.test.ts
@@ -1,0 +1,73 @@
+import * as THREE from 'three';
+import {describe, it, expect} from 'vitest';
+
+import {XRDeviceCamera} from './XRDeviceCamera';
+import {
+  getCameraParametersSnapshot,
+  isDeviceCameraPoseAvailable,
+} from './CameraUtils';
+
+function makeXrCameras(count: number): THREE.WebXRArrayCamera {
+  return {
+    cameras: new Array(count).fill({}),
+  } as unknown as THREE.WebXRArrayCamera;
+}
+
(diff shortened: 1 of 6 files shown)


dli7319:
It's working again! TY!

salmanmkc:
> It's working again! TY!

my pleasure! sorry about that!

PR #387, opened 2026-06-17
add magic_window: segment people out of the camera onto a swappable backdrop

salmanmkc:
New demo. uses MediaPipe ImageSegmenter (the multiclass selfie model, which wasn't wired up anywhere in the SDK yet) to cut people out of the device-camera feed in real time and drop them onto a swappable backdrop, shown on a floating window in the scene.

per frame it grabs a camera snapshot, runs the segmenter for a category mask, and composites in a shader: person pixels kept, background replaced with passthrough (see straight through to the room), a solid colour, or a gradient. the camera comes straight from the SDK camera module (`enableCamera('user')`), so the desktop simulator pulls the real front webcam and a headset uses its world-facing camera, no demo-side getSnapshot patching. the window sizes itself to the camera aspect so the feed isn't stretched.

A draggable uiblocks control panel swaps backdrops, same ManipulationBehavior card + renderOrder/depthTest treatment as the other demos so the labels aren't occluded by their backplate. keyboard B cycles them too.

prototyped demo-first; the segmenter wrapper (`SegmenterController`) is kept self-contained so it can be lifted into a `world/segmentation` addon later.

Passthrough should work on XR devices (the window frame is a ring now so the cut-out shows the real world through the middle, not a dark backplate).

updated passthrough:
updated passthrough

ignore the passthrough being black in the video below (older video)

https://github.com/user-attachments/assets/241e05d3-64af-41e6-8369-22c919df74ef


ruofeidu:
Love it! Thank you Salman!!

salmanmkc:
> Love it! Thank you Salman!!

my pleasure! thanks for the quick review!

demos/magic_window/WebcamFallback.js
```diff
@@ -0,0 +1,127 @@
+// Webcam fallback for desktop / simulator mode.
+//
+// The SDK reads frames from `xb.core.deviceCamera.getSnapshot()`, which on a
+// real Quest returns frames from the passthrough cameras. On desktop the
```
ruofeidu: nit: avoid using Quest in the comments, but XR device

demos/magic_window/WebcamFallback.js
```diff
@@ -0,0 +1,127 @@
...
+const VIDEO_H = 480;
+
+let _videoEl = null;
+let _canvasEl = null;
+let _canvasCtx = null;
+let _origGetSnapshot = null;
+
+function shouldUseWebcamFallback() {
+  // Only activate when the simulator addon is in charge (i.e. we're not in
+  // an immersive WebXR session). On a real Quest this returns false and the
```
ruofeidu: nit: avoid using Quest in the comments, but XR device

demos/magic_window/WebcamFallback.js
```diff
@@ -0,0 +1,127 @@
...
+// (no `immersive-ar` WebXR session active). Real-headset path is untouched
+// because the simulator addon never initialises when the device supports
+// WebXR.
+//
+// Permission flow: we call `getUserMedia()` directly. Chrome shows its
+// native permission prompt with the site URL and a one-click allow / deny.
+// If the user denies, the SDK keeps using the original `getSnapshot`
+// (which returns blank/scene frames) and the demo degrades to "no face
+// detected" without crashing.
+
```
ruofeidu: you can have this for now.
Yet I remember we can simply set simulator webcam in the camera module.

the next camera is kind of broken --- I'll talk with Nels in the afternoon.
https://xrblocks.github.io/docs/templates/Camera/

demos/magic_window/WebcamFallback.js
salmanmkc: sure will change it!

demos/magic_window/WebcamFallback.js
salmanmkc: will change!

demos/magic_window/WebcamFallback.js
salmanmkc: oo great thanks, looks like Nels did the fix https://github.com/google/xrblocks/pull/392 I will adjust it now

ruofeidu (approved):

PR #388, opened 2026-06-17
Embodied Control + Remote Control

whuang37:
Native APIs for embodied control. Control scheme is based off AI2Thor through gym-like step based control with multi action blocks per step (one for locomotion, rhand, and lhand) accordingly. New addon ```remote-control``` that streams out a websocket for embodied control to run multiple playwright instances.

Organized Changes:

```src/core/core.ts```
- added new options for pausing/stopping the physics and rendering loop to enable step by step replay.

```src/addons/embodied-control```
- Added new api input for embodied control of the user agent
- enables screenshot rendering as output
- realtime replay (every step is rendered as a set of frames) or low overhead step by step replays, rendering only individual steps

```src/addons/remote-control```
- lightweight wrapping over embodied-control that opens a local websocket for embodied control io

```samples/embodied_control```
- simple sample showing embodied control through sidebar options.

Example was created in samples following conventions of documentation examples. Can probably be moved to demos. 

Video demo of sample [here](https://youtu.be/BX-opLUNMo8)

### CODE REVIEW

File: src/core/Core.ts [hunk 1]
```diff
@@ -157,6 +160,32 @@ export class Core {
...
+  pause() {
+    this._isPaused = true;
+  }
+
+  resume() {
+    this._isPaused = false;
+  }
+
+  stepFrame(dtMs = 16.67) {
+    const wasSteppingFrame = this.isSteppingFrame;
```
dli7319: In what case would this be true already?

dli7319:
Are the "steps" and "controls" tied together? I.e. it's only applicable for testing?
Or can embodied control be used without having control over the frame loop?

dli7319 (approved):

src/core/Core.ts
whuang37: currently step + control are tied together for testing. we could have it be untied through something like embodied.applyControl(control) vs await embodied.step(durationMs, control) to enable both but i had trouble finding a reason for running this without the frame loop.

This would return true when stepping during an existing step ie during core.stepFrame() some script recalls stepFrame() as a reentrance guard due to nested stepping.

dli7319: It doesn't seem to guard right now. It just steps again anyways. Is this intended?

whuang37: That was a mistake. Should throw an exception. Added the apply control + step since it doesnt really change any logic just exposes a new method.

PR #389, opened 2026-06-17
docs: add 3D Object Boxes sample page

salmanmkc:
adds the objects_3d demo to the docs site so it shows up in the samples nav under AI + XR, next to the other gemini object samples. same SamplesIFrame page pattern as the rest, plus the sidebar entry. demo's already on main so the iframe loads once deployed.


ruofeidu:
Validated again and it works great

PR #391, opened 2026-06-17
Simulator Hand Glow

whuang37:
Following Android XR added a simple hand glow effect for hand visualization under occlusion. 

image

### DIFF
--- a/docs/docs/manual/Simulator.mdx
+++ b/docs/docs/manual/Simulator.mdx
@@ -69,10 +69,6 @@ Currently, the simulator supports the following features:
 
 - Depth simulation.
 - Hands simulation.
-- An optional simulator hand glow. When `simulator.renderToRenderTexture` is
-  enabled, visible simulator hands keep their normal appearance while an
-  always-on-top white glow is composited around their silhouette. Configure it
-  with `options.simulator.handGlow`.
 
 Notably, the following features are not supported.
 
--- a/docs/docs/manual/Simulator.mdx
+++ b/docs/docs/manual/Simulator.mdx
@@ -69,6 +69,10 @@ Currently, the simulator supports the following features:
 
 - Depth simulation.
 - Hands simulation.
+- An optional simulator hand glow. When `simulator.renderToRenderTexture` is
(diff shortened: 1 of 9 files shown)


ruofeidu:
Look good to me. 

@dli7319 , shall we make it default as this or an option to toggle, I'm fine.

maybe this could let agents understand deeper that this is "simulated user hand?"

dli7319:
In my opinion, this shouldn't be simulator specific since real hands will be occluded by virtual objects.
Can you refactor it into a its own `xb.Script` file under 'src/addons/hands/HandGlow.ts`?

dli7319:
Can we just stick it into the main scene with a high `renderOrder`?
I don't think we need to postprocess it in.

whuang37:
> In my opinion, this shouldn't be simulator specific since real hands will be occluded by virtual objects. Can you refactor it into a its own `xb.Script` file under 'src/addons/hands/HandGlow.ts`?

refactored into an addon. needed to add one more hook post render for it though which causes the core.ts test cases to fail. Could rewrite the test cases instead to handle this or leave the check for if afterrender exists as is. 

Instead of this addon, we could also just enable the simulator to have virtual object occlusion by rendering a sim hand proxy in the app layer which can then have occlusion with the app layer. That way the simulator hand acts like a physical hand in reference of the simulator scene

whuang37:
I've take a deeper look into this and im not exactly married to the concept of this being a feature. Since in existing webxr implementation, theres already a glow this addon would only really be used for the simulator since itd be the only way to show the hands under occlusion. Potentially a better path forward is a realistic hand occlusion toggle and then the hand glow/outline is only in the simulation as a representation of the hand tracked while the actual sim hand is a representation of the physical hand. Otherwise, I just don't foresee this being useful. @dli7319 @ruofeidu

whuang37:
not needed

dli7319:
I still do think we need an Android XR-like virtual hand for apps to integrate btw.

I just don't think it should be tied to the simulator or require post-processing. It should just be an independent scene graph object which is how it's implemented in the OS.

PR #393, opened 2026-06-18
Avatar performance fix

Peter-Pater:
Previously, running vrm avatars in XR was laggy. Fixed with the following improvements:

- Applied model optimization from VRMUtils, including redundant vertice removal, morph combination, and skeleton combination.
- Throttled the framerate for Springbones (Hair) to 30fps

Also disabled Grid. For some reason the grid contributed to the performance issue the most. When the grid was enabled, raycasting to the floor becomes very laggy.


ruofeidu:
LGTM!

ruofeidu (approved):

PR #394, opened 2026-06-18
Semantic embodied actions

whuang37:
Added new semantic embodied actions  in the ```embodied-actions``` addon for
- Teleporting to an object
- looking at an object
- pointing at an object (rotating the hand from where it is right now)
- reaching to an object (moving the hand to some object)
- clicking

### DIFF
--- a/src/addons/embodied-control/EmbodiedControlExecutor.ts
+++ b/src/addons/embodied-control/EmbodiedControlExecutor.ts
@@ -425,9 +425,11 @@ export class EmbodiedControlExecutor {
 
   async pointTo(
     handIndex: number,
-    target: THREE.Object3D | THREE.Vector3 | [number, number, number]
+    target: THREE.Object3D | THREE.Vector3 | [number, number, number],
+    options: {durationMs?: number} = {}
   ): Promise<EmbodiedControlStepResult> {
     return this.executeAction(async () => {
+      const {durationMs} = options;
       const {camera, simulator, core} = this.dependencies;
       const targetWorldPos = new THREE.Vector3();
       this.getTargetWorldPosition(target, targetWorldPos);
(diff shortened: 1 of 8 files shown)


dli7319 (approved):

src/addons/embodied-control/EmbodiedControlExecutor.ts
```diff
@@ -278,6 +280,304 @@ export class EmbodiedControlExecutor {
...
+    target: THREE.Object3D | THREE.Vector3 | [number, number, number],
+    options: {velocity?: number} = {}
+  ): Promise<EmbodiedControlStepResult> {
+    return this.executeAction(async () => {
+      const {velocity} = options;
+      const {camera, core} = this.dependencies;
+      const targetWorldPos = new THREE.Vector3();
+      this.getTargetWorldPosition(target, targetWorldPos);
+
+      if (velocity === undefined) {
```
ruofeidu: nit: will velocity ever be 0?

src/addons/embodied-control/EmbodiedControlExecutor.ts
```diff
@@ -278,6 +280,304 @@ export class EmbodiedControlExecutor {
...
+        camera.lookAt(targetWorldPos);
+        core.stepFrame(16.67);
```
ruofeidu: nit: magic number

ruofeidu (approved):

src/addons/embodied-control/EmbodiedControlExecutor.ts
whuang37: 0 or negative would break so added a guard there

src/addons/embodied-control/EmbodiedControlExecutor.ts
whuang37: moved to options as tickMs configurable

PR #398, opened 2026-06-20
world: add a segmentation subsystem (person/background masks)

salmanmkc:
magic_window's segmenter was just for the demo, so i pulled it up into the SDK as a proper `world/segmentation` block so that now any app can get person/background masks without rewiring MediaPipe itself.

`options.enableSegmentation()` then `await xb.core.world.segmentation.runSegmentation()` gives you a per-pixel category mask `{data, width, height}` (selfie multiclass, on-device, pulls frames through the camera module). mirrors `world/humans` and `world/faces`, just simpler since it's a pure 2d camera to mask, no depth or world-space step.

magic_window now uses it instead of its own `SegmenterController` (deleted), so the primitive's got a real consumer proving it end to end and the duplicated mediapipe wiring is gone.

`enableSegmentation()` turns the camera on (environment by default, same as `enableFaceDetection`); apps override facing if they want 'user' like magic_window does.

also flipped the demo's warm-up fallback, until the first mask lands it now shows the backdrop instead of briefly flashing the raw room.

colocated tests for the options defaults/merge, the category enum, and the mask-mapping helper.


demos/magic_window/MagicWindow.js
```diff
@@ -77,9 +75,10 @@ const FRAGMENT_SHADER = /* glsl */ `
     // The camera texture uses GL flipY; the mask DataTexture does not, so
     // flip the mask's v to line the two up.
     float id = texture2D(uMask, vec2(vUv.x, 1.0 - vUv.y)).r * 255.0;
-    // Until the first mask arrives, show the raw feed. Category 0 is the
-    // background; everything else is a person.
-    bool isPerson = (uHasMask < 0.5) || (id >= 0.5);
+    // Until the first mask arrives, treat the whole frame as background so we
+    // show the backdrop rather than flashing the raw camera (the room) during
+    // model warm-up. Category 0 is background; everything else is a person.
+    bool isPerson = (uHasMask >= 0.5) && (id >= 0.5);
```
ruofeidu: nit: > instead of >= ?

src/world/segmentation/SegmentationOptions.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    const returned = opts.enable();
+    expect(opts.enabled).toBe(true);
+    expect(returned).toBe(opts);
+  });
+
+  it('deep-merges constructor overrides while keeping unspecified defaults', () => {
+    const opts = new SegmentationOptions({
+      backendConfig: {
+        mediapipe: {
+          modelAssetPath: 'https://example.com/custom.tflite',
```
ruofeidu: AI-genereated?
If it's open-source asset, you can stage to xrblocks/assets etc.
Likely MediaPipe has a CORS one?

ruofeidu: Oh, I saw the one below, then this one should be some placeholder?

salmanmkc: yeah this is some placeholder, I could change it to model://custom instead, that is what other placeholders seem to be like

demos/magic_window/MagicWindow.js
salmanmkc: yeah good suggestion!

whuang37:
@salmanmkc working on realigning the world modules right now. Could you refactor the segmenter module into the same setup as world.HumanRecognizer, world.ObjectDetector, world.FaceRecognizer right now on main? It should just add a continuous loop + polling so that we can limit the number of repeated run detection calls when multiple modules need the same data. Thanks!

https://github.com/google/xrblocks/blob/main/src/world/humans/HumanRecognizer.ts as an example.

salmanmkc:
> @salmanmkc working on realigning the world modules right now. Could you refactor the segmenter module into the same setup as world.HumanRecognizer, world.ObjectDetector, world.FaceRecognizer right now on main? It should just add a continuous loop + polling so that we can limit the number of repeated run detection calls when multiple modules need the same data. Thanks!
> 
> https://github.com/google/xrblocks/blob/main/src/world/humans/HumanRecognizer.ts as an example.

sure! will do that, cool stuff!

salmanmkc:
I think this should match the same setup now

whuang37 (approved):
Has the functionality i was looking for. Thanks!

ruofeidu:
Screenshot 2026-06-26 at 3 36 22 PM
How can I trigger the segmentation?

Ah, we likely need a real person asset to add to the simulator view... We can still use in XR, but it would be confusing for desktop testers.

ruofeidu:
My bad, I had a physical camera occluder :) 

LGTM

ruofeidu:
There might be double GPU-to-CPU Canvas Readback issues. 

Yet it is not a big problem for initial PR.

```
let MagicWindow display the camera texture directly (using xb.core.deviceCamera.texture) rather than converting the snapshot back and forth from ImageData to a CanvasTexture.
```

we have a demo here long time ago: https://storage.googleapis.com/tfjs-models/demos/relighting/index.html
or code here: https://github.com/tensorflow/tfjs-models/tree/master/depth-estimation/demos

ruofeidu (approved):

PR #402, opened 2026-06-22
Added a Squeeze Action

whuang37:
Fixed the squeeze hookup and added it as an embodied action.

In ```embodied-action``` makes a fist to squeeze. 

Unified a ```release``` high level action which clears the state of the hand entirely.

Fixed some small indexing bugs.

### DIFF
--- a/docs/docs/manual/EmbodiedControl.mdx
+++ b/docs/docs/manual/EmbodiedControl.mdx
@@ -84,10 +84,7 @@ the camera. Locomotion `rotate` is `[pitchDegrees, yawDegrees, rollDegrees]`.
 Hand `move` and `rotate` are relative to the simulator controller pose. Hand
 `rotations` can apply sparse simulator joint rotations in radians.
 
-Use `selectStart` and `selectEnd` for pinch-like primary actions. In the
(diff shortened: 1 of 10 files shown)


dli7319:
I'm not aware the squeeze can be triggered by a fist gesture. Could you check this on device?

ruofeidu:
As far as I remember, Android XR doesn't have a native support using hands without controller.
So this is lower priority, but a nice thing to fill in the gap.

For now, we don't want to encourage usage of `onSqueeze`, since many AXR users won't have controllers (they are optional), but thanks for the PR.

dli7319:
IMO It's fine to add `onSqueeze` but make it not able to be triggered with hands and add a disclaimer comment that it's for controller only.

whuang37:
Gotcha thats what I kind of thought but when referring to webxr docs it states "hand gesture of grabbing" 

https://developer.mozilla.org/en-US/docs/Web/API/XRSession/squeezestart_event

I think its pretty small either way since i was just doing this to fill in the final gaps of all the missing pieces for embodied actions. If we do just for controller its probably pragmatic to not merge this since its pretty much only hand -> on squeeze and leave the original as is with a new pr for just the hand indexing fix

PR #405, opened 2026-06-24
Generative objects: summon AI-generated objects onto the surfaces around you

salmanmkc:
## Description

opening `demos/generative_object/` against the current SDK stopped before startup. Clear could bring a cancelled object back when Gemini responded, removing the demo left objects and controls alive, and entering a key reloaded the page with that key in the request URL.

ports the demo to built-in spatial UI and `xb.manipulation`, aligns three.js with 0.185.0, and keeps prompt-to-object generation demo-owned. buttons or voice request an image, remove its background and place the cutout on a nearby depth surface, with an in-front fallback. the SDK additions are pure image-processing and placement helpers only, not a Core/Options generative subsystem. optional 2.5D relief still uses brightness as an approximation, not measured depth.

Clear and teardown now invalidate work before the AI request and after image decoding. old completions cannot place objects or overwrite a replacement demo's status. both scripts release owned objects, speech handlers, DOM controls and lights; UI is released through the normal lifecycle. each cleanup step runs independently if another release throws.

the global depth-raycast override was deleted, not restored during teardown. current pointer filtering already ignores the depth mesh, while direct placement raycasts still hit it. occlusion requires both depth and occlusion to be enabled, including when a disabled Depth is registered. each object tracks every compiled shader variant and removes all registrations on disposal.

key entry now passes through `options.ai.gemini.apiKey`, without navigation or adding it to the URL. legacy URL input is consumed and stripped before resource loading, and referrers are suppressed. the overlay now says plainly that the initial URL-key request may already have been logged. demo-local and repository-root `keys.json` files use the SDK's `{gemini: {apiKey: "..."}}` shape, and key resolution finishes before initialization. localStorage and client-side keys remain local-prototype conveniences only; production apps must proxy AI calls through a server they control.

### Coverage

red-first evidence is against #405 after merging `c8d86793`, not against unmodified main, because main has no generative demo. generation/ownership tests exposed the old behavior with a temporary legacy-enum shim to get past its removed import. UI teardown and stale-status tests were red after the API port but before the lifecycle fixes. real Chromium also reproduced the unchanged overlay ignoring a valid SDK keys file and navigating with the typed key.

preservation coverage is separate: the existing image-processing, sizing and facing helpers, already-correct cancellation during image decoding, distinct-texture disposal, current semantic UI composition, and direct depth placement while shared pointer hits stay disabled. there were 19 helper tests before the repair, not 45; 31 demo tests were added, for 50 focused tests total. the SDK build, full suite, lint, formatting and demo/test type checks passed locally.

### Browser checks and limits

Chromium loaded the freshly built SDK and compiled demo. manual, stored, local-file, root-file and legacy URL keys reached the expected options without another navigation or subsequent key-bearing request/referrer. fixture-image decoding and generation, Clear during an AI wait, and removal/recreation of both scripts passed without module exceptions. live Gemini requests were blocked during the smoke test.

spatial text is fragmented in this Chromium run. the same failure reproduces in untouched main's `templates/01_spatial_ui/` with depth disabled and in a minimal public UICard, so it is not demo-specific. the SDK-versus-headless-renderer cause is not isolated, and visual sign-off remains incomplete. no unrelated SDK UI changes were made.

no headset, real hand/controller dragging, live AI output quality or microphone hardware/permission behavior was verified. keeping this draft.

## Type of Change

- [x] Bug fix
- [x] New demo or sample
- [x] Documentation update

- **Simulator Recording**:  Not attached. Chromium smoke results and the spatial-text limitation are described above.
- **Device Recording**:  Not available; no device testing.

## Checklist

- [x] **Tested in simulator & device**: Verified functionality in desktop simulator and/or physical hardware (where applicable). Chromium desktop only, with fixture AI and the limitations above; no physical device verification.
- [x] **Large Assets ($\ge$ 1MB)**: Submitted separately to [xrblocks/proprietary-assets](https://github.com/xrblocks/proprietary-assets) via jsdelivr CDN. N/A: no assets of this size added.
- [x] **SDK Dynamic Dependencies**: All new SDK dependencies are dynamically loaded at runtime. N/A: no new SDK dependencies; three.js remains external.


ruofeidu:
Hi Salman,

Thank you for your contribution in this!!! I would like to request to switch to a demo.
Also the demo ignores `keys.json` file I used to debug locally.

I won't say this is ready to be put inside the SDK (for now).
Even for inside SDK, it should be under `ai.generateBillboard(image), etc.`

We need to carefully think of the high-level picture of `generativeAssets`:

abstract --------> photorealistic
photo vs. mesh vs. 3D Gaussians etc.
LLM / local model vs. cloud models

Internally, we have a demo like this, but with better quality & confidential tech :)

salmanmkc:
> Hi Salman,
> 
> Thank you for your contribution in this!!! I would like to request to switch to a demo. Also the demo ignores `keys.json` file I used to debug locally.
> 
> I won't say this is ready to be put inside the SDK (for now). Even for inside SDK, it should be under `ai.generateBillboard(image), etc.`
> 
> We need to carefully think of the high-level picture of `generativeAssets`:
> 
> abstract --------> photorealistic photo vs. mesh vs. 3D Gaussians etc. LLM / local model vs. cloud models
> 
> Internally, we have a demo like this, but with better quality & confidential tech :)

you sure the keys.json wasn't in the repo root or the wrong folder? should be relative 

and sure will move this to demo only, wow that's cool to know there's a better internal demo, is it sorta similar to likeness level quality?

ruofeidu (approved):

ruofeidu (changes requested):
upgrade this to the latest SDK and see if it still works, thanks for the update

PR #407, opened 2026-06-24
Rescoping Embodied Control and Remote Control

whuang37:
Refactored embodied control and remote-control to be a bit easier for agent control.

```embodied-control```
- removed all the observation handling to keep it purely action based
- fixed a small bug where autopause would pause before the world is loaded

```remote-control```
- refactored to use a small request/response protocol via ```ping``` and ```callTool``
- now all observation and actions are wrapped in a tools which can be easily extendible depending on the remotecontrol injection needs.
- ```RemoteControlClient``` wraps callTool() for common actions and observations
- localhost relay via ```src/addons/remote-control/server/relay.js

```remote-control``` architecture: External Agent -> WebSocket relay -> browser page with RemoteControl -> tools -> JSON response
- uses a relay to allow python/js clients to connect without enforcing anything

Usage
1. Inject the remote control into some scene startup file
```
import * as xb from 'xrblocks';
import {RemoteControl} from 'xrblocks/addons/remote-control/index.js';

const app = new MyScene();
xb.add(app);

const options = RemoteControl.configureOptions(new xb.Options());

xb.add(
  new RemoteControl({
    url: 'ws://127.0.0.1:8791',
    reconnect: true,
    embodiedOptions: {autoPause: true, realTime: false},
    tools: {
      getSceneState: async () => app.getState?.(),
      resetScene: async () => app.reset?.(),
      performSceneAction: async (args) => app.performAction?.(args),
    },
  })
);

await xb.init(options);
```

2. Startup the relay ```npx xrblocks-remote-control```
3. Open the web page e.g.: ```http://127.0.0.1:8080/samples/remote_control/?remoteControlSession=run-0```
4. Send and receive data via a websockets client

Demo [here](https://youtu.be/t-9iNaLn3Fs) showing running embodied actions via a websocket

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -80,6 +83,9 @@
   "peerDependencies": {
     "three": "^0.184.0"
   },
+  "optionalDependencies": {
+    "ws": "^8.18.3"
```
dli7319: Can you move this to dev dependencies?

dli7319 (approved):
Great work!
Now we don't have to inject js calls into playwright sessions?

whuang37:
> Great work! Now we don't have to inject js calls into playwright sessions?

Yep! Unfortunately theres no way I could think of to avoid having to do a bit of code injection to get any sort of scene information beyond keyboard controls but it keeps things more minimal and lets us do things a bit easier

package.json
whuang37: moved!

PR #410, opened 2026-06-25
Remove debug object detection png download.

dli7319:


salmanmkc (approved):

ruofeidu:
Sorry this is an intended feature.

Initially I mean by setting `this.context.options.objects.showDebugVisualizations = false` for our readers. In second thoughts, maybe we just leave it there.

PR #411, opened 2026-06-25
ui: Unify panel corner radius and fix edge dragging

dli7319:
- Added `canDrag(intersection)` optional hook on the `Draggable` interface.
- Integrated `canDrag` in `DragManager.beginDragging` to prevent dragging from panel center, child elements, or non-border regions.
- Replaced percentage-based `panelScale` with absolute `borderWidth` (in meters) configurable via `PanelOptions`.
- Implemented `Panel.canDrag` using physical borders in UV space.
- Modified `PanelMesh` to use a unit square `PlaneGeometry(1, 1)` and scaled it dynamically in `Panel.updateLayout`.
- Updated `SquircleShader` to use absolute physical dimensions and `fwidth` for resolution-independent antialiasing, unifying corner radius with `SpatialPanelShader`.
- Fixed child panel relative sizes and weights in `samples/ui/index.html` to prevent overflow of parent borders.
- Added comprehensive unit tests in `src/ux/DragManager.test.ts` for all dragging edge cases.

image

### DIFF
--- a/samples/ui/index.html
+++ b/samples/ui/index.html
@@ -174,7 +174,7 @@
           const panel = new xb.SpatialPanel({
             width: 1.3,
             height: 1.25,
-            backgroundColor: '#00000000',
+            backgroundColor: '#21252baa',
           });
           panel.position.set(
             0,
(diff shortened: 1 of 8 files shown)


ruofeidu:
Open disucss: I think we are breaking a behavior where panels were set draggable anywhere by default.

Initially this was requested by Ben, but maybe we should make it an option: drag by border or drag by panel by default?

dli7319:
Let me split this PR up then. I want to fix the edge sizing.

ruofeidu (approved):

PR #414, opened 2026-06-25
ui: Unify panel corner radius and use physical border width

dli7319:
Instead of making spatial panels 1.3x the size of normal panels, make them `borderWidth` larger.

image

### DIFF
--- a/samples/ui/index.html
+++ b/samples/ui/index.html
@@ -190,8 +190,8 @@
           // player row
           const playerRow = grid.addRow({weight: 1.0});
           const playerPanel = playerRow.addPanel({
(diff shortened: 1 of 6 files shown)


ruofeidu:
borderWidth makes more sense!

The indent of the photo gallery looks still small, maybe also for other panels, and we can set a heuristics for floating panels.

ruofeidu (approved):

dli7319:
Photogallery goes up to the border along the top and bottom.
The border is part of the panel for the bottom three panels but not the top panel.
I don't like this inconsistency btw.

PR #416, opened 2026-06-26
feat(addons/agenthands): hands + orb that gesture and point as the agent speaks

salmanmkc:
adds a free-standing pair of agent hands and a glowing orb that gesture while the agent talks and physically point at real things in the room. inspired by the AgentHands paper (CHI 2026, https://www.duruofei.com/papers/Liu_AgentHands-GeneratingInteractiveHandsGesturesForSpatiallyGroundedAgentConversationsInXR_CHI2026.pdf).

Demo (youtube link: https://youtu.be/JgDE5crT3MQ)

https://github.com/user-attachments/assets/1fd35112-970e-4120-b75d-283bf8daec9c

the paper's nice insight is that an agent feels a lot more present when it can gesture and actually point at things in your space, not just talk at you. It lays out a clean way to drive that from an llm (inline gesture markup), grounds the gestures to real objects, and keeps the embodiment deliberately minimal: a calm orb as the locus of attention plus translucent hands, no face, so it doesn't tip into uncanny. 

Without a key the demo plays a short scripted monologue so you can see the gestures. add `?key=...` and it's the full loop: you talk (speech recognizer), gemini replies with inline gesture markup, the reply is spoken (TTS) and the hands gesture in sync with the actual spoken words, pointing at real detected objects when it refers to them. runs on desktop through the simulator, and in headset via a small spatial control panel (talk / scan).

new addon `src/addons/agenthands`:

- `AgentHand`: one posable hand rig. loads the webxr generic-hand glb, poses the bones toward a `SimulatorHandPose`, and can aim its index finger at a world point (pivoting from the wrist so close-range pointing stays accurate). a layered motion offset lets gestures move the hand on top of the pose and aim.
- `AgentHands`: the left + right pair. dispatches gestures and points, picks the pointing hand by which side the target is on, and has beat / wave / iconic size / count motions.
- `AgentGestures`: parses inline markup from the reply, `[gesture:thumbs_up]`, `[point:the lamp]`, `[wave]`, `[beat]`, `[size:big]`, `[count:2]`, into poses, motions and point targets, anchored to the spoken text so each fires on the right word.
- `AgentHead`: the agent's presence, a semi-transparent blue orb with a drifting particle shell. breathes while idle, pulses while speaking, gazes at whatever it's pointing at. matches the paper's embodiment (orb as the locus of attention, no face), and the hands are the same translucent blue.

one SDK change: `SpeechSynthesizer.onBoundaryCallback`, fired on each word boundary with the character index into the spoken text, so callers can sync visuals (here, gestures) to the actual spoken words. optional, off by default.

pointing is grounded with a lightweight raycast against the depth mesh, so the demo only needs gemini and no extra detection deps. one thing that ended up different from the paper: there, you register objects one at a time in a dedicated mode (look at a thing, say "register this", and it builds a rich oriented box with face and region labels that lands in a registry). Here the agent detects the whole room in a single pass and keeps re-grounding in the background as you move, so the grounding isn't a one-time observation, it stays current as you walk around without any manual step. the tradeoff is that it's coarser: a point per object rather than the paper's region-level boxes. the richer 3D object-detection addon that grew alongside this (`objects3d`) is going up as its own PR.

Why bring it into xrblocks: the paper's system is a unity study app on a galaxy xr headset. this ports the idea to the open web on three.js / webxr, packaged as a reusable addon so it's something you can drop into an app rather than a one-off, and it runs in the desktop simulator so you can try the whole loop without a headset. the gesture taxonomy, the markup-driven control, and the minimal orb-plus-hands embodiment all follow the paper; the main new plumbing is syncing gestures to the spoken words through the small `SpeechSynthesizer` hook above.

Worth being upfront the per-hand gesture state machine is simpler, there are fewer gesture types, and timing comes from tts word boundaries rather than the paper's per-word energy model. pointing leans on depth-mesh quality and 2D detection instead of a full scene mesh, and there's no user-gaze input yet (the orb gazes, but we don't read where you look). most of the tuning happened in the simulator, so the in-headset path is wired but less exercised. plenty of room to grow it (more gestures, gaze, better grounding), which is part of why it's an addon.

colocated vitest specs for the addon cover the hand rig, the pair, the gesture parser and the orb. lint, prettier and build are clean.


ruofeidu:
Thank you Salman for the contribution! Do you have a recording of this?

I'm just back from a trip recently and a bit slow catching up.

salmanmkc:
> Thank you Salman for the contribution! Do you have a recording of this?
> 
> I'm just back from a trip recently and a bit slow catching up.

I will get one done for you soon! Hope you had a good trip!

salmanmkc:
> Thank you Salman for the contribution! Do you have a recording of this?
> 
> I'm just back from a trip recently and a bit slow catching up.

Added a video to the PR description, some future improvements we need e.g. moving the head etc. but for now I think it's a good state. Can also watch it here: https://youtu.be/JgDE5crT3MQ

https://github.com/user-attachments/assets/840bfa9d-366a-494c-8354-4ec5bdc50ed3

qxziuan (commented):
Hi salmanmkc, this is an awesome demo!

One high-level suggestion regarding main.js:

I read that AgentHands consists of four major modules: (1) World Understanding: Returns 3D object positions and saves them to local storage. (2) Clip-Embedded Response Parser: Parses the AI output into an executable dictionary. (3) TTS Timestamp Matcher: Tracks which word is currently being spoken. (4) Gesture Animator: Controls the AgentHands to change gestures and move.

I see you have all the necessary functions for these inside main.js, but I highly recommend refactoring these key modules out into src/addons/agenthands/ to keep the code modular and easier to maintain.

Also, it would be great to add a quick README that credits the AgentHands paper and clarifies what this demo can and can't do right now.

salmanmkc:
> Hi salmanmkc, this is an awesome demo!
> 
> One high-level suggestion regarding main.js:
> 
> I read that AgentHands consists of four major modules: (1) World Understanding: Returns 3D object positions and saves them to local storage. (2) Clip-Embedded Response Parser: Parses the AI output into an executable dictionary. (3) TTS Timestamp Matcher: Tracks which word is currently being spoken. (4) Gesture Animator: Controls the AgentHands to change gestures and move.
> 
> I see you have all the necessary functions for these inside main.js, but I highly recommend refactoring these key modules out into src/addons/agenthands/ to keep the code modular and easier to maintain.
> 
> Also, it would be great to add a quick README that credits the AgentHands paper and clarifies what this demo can and can't do right now.

Thanks so much @qxziuan! I've done as you suggested, pulled all four out of `main.js` into `src/addons/agenthands/`.

1. world understanding -> `AgentWorld`: detection + depth-mesh grounding + the object cache, and it persists to localStorage like you mentioned.
2. clip-embedded response parser -> `buildGestureSteps` in `AgentGestures`: markup into a timed step list, with each point target resolved to a world position.
3. tts timestamp matcher -> `AgentSpeechConductor`: drives the timeline and syncs it to the synth's word boundaries.
4. gesture animator -> `AgentGestureAnimator`: poses / motions / pointing, and tracks which hand is pointing for the pointer viz and the orb's gaze.

each has colocated vitest specs. while wiring it up i also caught a latent double-play: the word-boundary sync and the timed queue could both fire the same motion, so a step now plays once per utterance.

Yeah that's a good idea, I tried to credit in this PR description, but putting in in a readme is a good idea, I've added 
 `demos/agent_hands/README.md`. it credits the paper and is upfront about what the demo does and doesn't do vs it yet, gesture range is a smaller flat set (no two-hand iconic, affective vfx, measurement, or high-five), timing comes off the tts word boundaries rather than a per-word energy model, grounding is one depth-mesh point per object instead of the gaze-registration flow with oriented boxes, no user-gaze input, and the hands stay anchored in front of you rather than the agent walking out to the target. 

There's also a short `src/addons/agenthands/README.md` for the addon api, and an `xb-agenthands` skill.md file that can aid agents, based on the pattern @ruofeidu made before.

skills/xb-agenthands/SKILL.md
qxziuan, line 1: @ruofeidu @dli7319 Need a sanity check on this new skills. Do we add them? And how do we gate this feature properly?

qxziuan (commented):
Thanks for the major revision. I left a comment in the SKILL.md to ask other contributors to give it a review.

skills/xb-agenthands/SKILL.md
dli7319: The feature is already separated into an addon. I think it's fine to submit.

dli7319:
Really awesome work! Great to see this finally implemented in XR Blocks!
I have some minor suggestions but I'm happy to merge it as-is.

1. The panel is largely outside of the FOV on my Galaxy XR. On device, it needs to be dragged a bit further.
2. Slightly weird that the agent is soft head locked. Also something looks wrong with the hands transparency (maybe double sided).
3. The voice from `SpeechSynthesizer` sounds unnatural. IIRC the Unity implementation used Gemini voices, but this might not be integrated yet into XR Blocks.
4. The FPS is low (i.e. the panel jitters as we move around) but I think that isn't specific to this demo. We'll probably need to do another performance audit on Galaxy XR.
5. "Thinking" takes a while these days. Most likely due to the Gemini API being overloaded during working hours.

dli7319 (approved):

salmanmkc:
> Really awesome work! Great to see this finally implemented in XR Blocks! I have some minor suggestions but I'm happy to merge it as-is.
> 
> 1. The panel is largely outside of the FOV on my Galaxy XR. On device, it needs to be dragged a bit further.
> 2. Slightly weird that the agent is soft head locked. Also something looks wrong with the hands transparency (maybe double sided).
> 3. The voice from `SpeechSynthesizer` sounds unnatural. IIRC the Unity implementation used Gemini voices, but this might not be integrated yet into XR Blocks.
> 4. The FPS is low (i.e. the panel jitters as we move around) but I think that isn't specific to this demo. We'll probably need to do another performance audit on Galaxy XR.
> 5. "Thinking" takes a while these days. Most likely due to the Gemini API being overloaded during working hours.

Thanks so much, and thanks for testing it on an actual Galaxy XR.

Since you're happy to merge as-is, I'd keep this PR focused and handle the panel placement, head anchoring, and hand material in a follow-up device-polish PR.

I did try the higher-quality Gemini voice path earlier, but the gesture timing depends on reliable word-boundary timestamps and that path didn't provide them, so the hands drifted out of sync. The browser voice sounds less natural, but it gives us character boundaries to keep the gestures lined up. Getting Gemini voices back would need a separate alignment path, likely forced alignment or an audio-based approach similar to the paper.

Agree the FPS and panel jitter need a broader Galaxy XR performance pass rather than something specific to this demo.

salmanmkc:
> Thanks for the major revision. I left a comment in the SKILL.md to ask other contributors to give it a review.

thank you!

salmanmkc:
I'll push small changes to resolve the branch conflicts 
CleanShot 2026-07-13 at 18 16 39

qxziuan (approved):
LGTM. Looking forward to the follow-up PRs.

salmanmkc:
> LGTM. Looking forward to the follow-up PRs.

thank you! just addressed the merge conflicts

salmanmkc:
@dli7319 picking up the soft head-locking follow-up from your feedback here. What behavior would you prefer for the hands + orb: stay in a fixed spot in the room with a recenter control, follow the user as they walk but ignore head turns, or only follow once they move far away?

The panel placement is already covered by #444, so this would just change how the agent itself is anchored.

PR #417, opened 2026-06-26
feat(addons/objects3d): reusable 3D object detection (2D detect + depth into oriented boxes)

salmanmkc:
adds an `objects3d` addon that turns 2D detection plus the depth mesh into oriented 3D bounding boxes you can query in world space. it's the pipeline that lived inline in the `objects_3d` demo, pulled out into a reusable `Script` so apps and other addons can just ask for detected 3D objects.

`Object3DDetector` runs the whole thing: snap the camera and depth mesh, run 2D detection, get a per-object segmentation mask, raycast the masked depth samples into world space, fit a yaw-aligned oriented box, fuse across views, and optionally draw debug wireframe boxes. each result is a `Detected3DObject` with a nearest-surface query, so you can ask for the closest point on an object, which is handy for pointing at it or placing something against it.

the backends are pluggable:

- 2D detection: `gemini` (open-vocabulary, needs a key) or `mediapipe` (on-device COCO, no key, fixed class set).
- masks: SAM / slimsam (via `@huggingface/transformers`) or the mediapipe segmenter.

it also exports the pure helpers that are useful on their own: `uvToNdc` (the snapshot-vs-camera aspect correction), `box2dIoU` / `unionDetections` / `snapBoxToFloor` (2D fusion and floor snapping), and the label `categorize` helpers (flat / surface / tiny-flat / light buckets).

the heavy deps (`@huggingface/transformers`, mediapipe) stay external; the one rollup change just marks transformers external so it isn't bundled.

this is split out of the agenthands branch (#416), which originally grounded its pointing through this pipeline before I switched it to a lighter depth raycast. the addon ships standalone here with colocated vitest specs (label categorization, depth sampling, fusion, and the nearest-surface query); the `objects_3d` demo can move onto it as a follow-up. lint, tests and build are clean.


src/addons/objects3d/geometry/Fusion.ts
```diff
@@ -0,0 +1,236 @@
...
+ *   {@link Detected3DObject}).
+ * @param newObb - Candidate OBB to merge.
+ * @param cat - Category of the candidate.
+ * @returns The matched record (already mutated) when fusion happened, or
+ *   `null` when no match was found.
+ */
+export function fuseIntoBoxes(
+  records: FusionRecord[],
+  newObb: InternalObb,
+  cat: string
```
ruofeidu: nit: category instead of cat

src/addons/objects3d/geometry/Fusion.ts
```diff
@@ -0,0 +1,236 @@
...
+): FusionRecord | null {
+  const half = (s: THREE.Vector3) => (s.x + s.y + s.z) / 6;
+  for (const rec of records) {
+    if (rec.category !== cat) continue;
```
ruofeidu: question:  will a chair and a sofa near each other (both furniture) fuse into one object?

src/addons/objects3d/Object3DDetector.ts
```diff
@@ -0,0 +1,530 @@
...
+          });
+          if (!obb) return null;
+
+          if (cat === 'furniture' || cat === 'small') {
+            snapBoxToFloor(obb, floorY);
+          }
+
+          const c = obb.center;
+          const farFromRoom =
+            Math.abs(c.x) > 6 || Math.abs(c.z) > 6 || c.y < -1 || c.y > 5;
```
ruofeidu: nit: declare magic numbers

ruofeidu (changes requested):
Some comments in line.

High-level: after tested on device --- Nels: can you give it a try?

consider runDetection({snapshot, backend}) override in ObjectDetector,
or before this is stable, it can live in addons and enable only when reqeusted.

src/addons/objects3d/geometry/Fusion.ts
salmanmkc: renamed, thanks!

src/addons/objects3d/geometry/Fusion.ts
salmanmkc: Yes, previously they could fuse because matching used only the broad category and center distance. I changed this so same-label observations retain the existing proximity matching, while differently labelled observations fuse only when both OBB centers lie inside each other's box. So tested it, it still handles one object changing from `armchair` to `sofa` across views, but keeps merely adjacent furniture separate. I added regression tests for both cases.

src/addons/objects3d/Object3DDetector.ts
salmanmkc: Thanks, pulled these into named constants, split by purpose: depth-sampling settings, room bounds, minimum samples per category, and fitted-box size and degeneracy limits.

salmanmkc:
> Some comments in line.
> 
> High-level: after tested on device --- Nels: can you give it a try?
> 
> consider runDetection({snapshot, backend}) override in ObjectDetector, or before this is stable, it can live in addons and enable only when reqeusted.

runDetection({snapshot, backend}) override in ObjectDetector as a follow up?

nsalminen (approved):

salmanmkc:
thanks @nsalminen 🥳

nsalminen:
> thanks @nsalminen 🥳

Thank *you* @salmanmkc! Apologies for the long review time. It was pending some internal discussions/decisions. Thanks for your work and patience!

salmanmkc:
> > thanks @nsalminen 🥳
> 
> Thank _you_ @salmanmkc! Apologies for the long review time. It was pending some internal discussions/decisions. Thanks for your work and patience!

No worries at all! Thanks for all the work behind the scenes! Very much appreciated @nsalminen to you and the team!

PR #418, opened 2026-06-26
ci: split tests into their own workflow

salmanmkc:
tests already run in ci but they're buried in build-check.yml as a `Run tests` step after the build. so if the build fails the tests never run, and a test failure just shows up under the "Build Check" check with no separate signal.

pulled them into their own `test.yml` ("Test"). same node 24 + npm cache setup as the others, runs `npm run test`. build-check now just builds, and i moved the `vitest.config.ts` path trigger over to the test workflow where it belongs.

`pull_request:` has no path filter so Test runs on every PR push, same as the existing workflows.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Build Check
+name: Test
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


salmanmkc:
Would be good to mark tests as required in settings after this pr is merged in (if you think it's a good change)

dli7319:
Great idea. Thanks Salman!

dli7319 (approved):

PR #419, opened 2026-06-26
feat(camera): read camera intrinsics + pose off the WebXR view (table stays fallback)

salmanmkc:
follow-up to #268. the device table still handles getUserMedia, but when WebXR Raw Camera Access supplies an image we can use that view's projection and pose instead of guessing calibration from the user agent.

captures the first usable camera image with that same view's matrices. the stored pose is reference-from-view; the world-space getter applies the render camera parent's updated world transform, including translated and rotated rigs. PlanarVST now includes both principal-point offsets, and camera parameter snapshots copy the projection so later frames cannot change half of a snapshot.

live parameters clear on source and session transitions, missing raw frames, fallback timeout and disposal. switching back to getUserMedia restores the video texture too. merged current main first, keeping its simulator camera interface and disposal guards.

regression tests failed before the fixes for translated/rotated rigs, asymmetric planar projection, mutable snapshots, raw-to-media transitions, session restart, missing frame data, fallback restart and disposal. preservation coverage checks first-usable-view selection, matching texture/matrices, copied WebXR arrays, symmetric frusta, the device table and simulator crop.

validated with the SDK build, full test suite (134 files, 1,870 passed and 1 skipped), lint and Prettier on every touched file. not hardware-verified. Raw Camera Access still needs device/browser support and permission; this does not make it available on Galaxy XR or Aura.


salmanmkc:
this only kicks in on the webxr raw-camera path and leaves the  DEVICE_CAMERA_PARAMETERS  table as the  │ fallback, so the getUserMedia path and the simulator are unchanged. i haven't been able to validate the numbers on  │ real hardware yet though, so it'd be good to confirm the intrinsics + pose line up on a galaxy xr and an aura before  │ leaning on it. those two share the android xr user agent, which is the whole reason for reading off the view instead  of the device table.

dli7319:
I haven't tested, but I suspect it likely won't work on Galaxy XR or Aura since it's not exposed at the Android or C++ api level afaik.

ruofeidu:
let me know when this is ready for reviews.

salmanmkc:
> let me know when this is ready for reviews.

Technically it's ready for review, if it doesn't work with pass through on aura or galaxy XR it will fallback safely

dli7319:
I checked on my Galaxy XR and WebXR Raw Camera Access API isn't available.

Screenshot 2026-07-31 at 4 49 06 PM


Is there a device that actually supports it?
Otherwise, this would be unreachable and untested code.

PR #425, opened 2026-07-02
core: request world-sensing features as optional instead of required

gagarinyury:
## Problem

When `options.world.planes`, `options.world.meshes`, or `options.lighting` are enabled, `Core` pushes `plane-detection` / `mesh-detection` / `light-estimation` into `requiredFeatures` of the immersive session request.

On browsers that do not support one of these features (e.g. `light-estimation` on standalone headset browsers, or `plane-detection` behind a flag on some devices), `requestSession('immersive-ar')` is rejected entirely — the Enter XR button silently does nothing, even though the app could run fine without the feature.

## Change

Request these three world-sensing features via `optionalFeatures` instead, following the existing pattern used for `camera-access` a few lines above.

No behavior change on devices that support the features. On devices that don't, the session now starts and the subsystems degrade gracefully — they already handle absent data:

- `PlaneDetector.update()` bails out when `frame.detectedPlanes` is missing
- `MeshDetector` bails out when the frame mesh collection is missing
- `Lighting` (`XREstimatedLight`) simply never receives `estimationstart`

`depth-sensing`, `local-floor`, and `hand-tracking` intentionally stay required — apps enabling them typically cannot run meaningfully without them.

## Testing

- `npm run build` and `tsc --noEmit` are clean
- `npm test` matches `main` (the 13 pre-existing failures in the gamepad-bindings suite are unrelated to this change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/core/Core.ts
+++ b/src/core/Core.ts
@@ -332,14 +332,13 @@ export class Core {
     }
 
     const webXRRequiredFeatures: string[] = options.webxrRequiredFeatures;
+    const webXROptionalFeatures: string[] = options.webxrOptionalFeatures;
     // Use camera-access when the browser supports it.
     if (options.deviceCamera?.enabled) {
-      if (!this.webXRSettings.optionalFeatures) {
-        this.webXRSettings.optionalFeatures = [];
-      }
-      (this.webXRSettings.optionalFeatures as string[]).push('camera-access');
+      webXROptionalFeatures.push('camera-access');
     }
     this.webXRSettings.requiredFeatures = webXRRequiredFeatures;
+    this.webXRSettings.optionalFeatures = webXROptionalFeatures;
     // Sets up depth.
     if (options.depth.enabled) {
       webXRRequiredFeatures.push('depth-sensing');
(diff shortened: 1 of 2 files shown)


gagarinyury:
@googlebot I signed it!

src/core/Core.ts
```diff
@@ -340,6 +340,18 @@ export class Core {
...
     }
     this.webXRSettings.requiredFeatures = webXRRequiredFeatures;
+    // World-sensing features below are requested as OPTIONAL: devices that
+    // lack them (e.g. light-estimation or plane-detection on some headset
+    // browsers) would otherwise reject the whole immersive session instead of
+    // running without the feature. The detectors already no-op when the
+    // session does not provide the data (PlaneDetector/MeshDetector bail out
+    // on a missing frame collection; XREstimatedLight simply never fires
+    // `estimationstart`).
+    if (!this.webXRSettings.optionalFeatures) {
```
dli7319: I think we should just have a new field `options.webxrOptionalFeatures`.
Then we can remove this if block and the identical if block above it.

dli7319 (changes requested):
Thanks for catching this!

src/core/Core.ts
dli7319: Similar to `options.webxrRequiredFeatures`.

gagarinyury: Done — added a `webxrOptionalFeatures` option mirroring `webxrRequiredFeatures`, and removed both `optionalFeatures` init guards. Thanks!

dli7319 (approved):

PR #426, opened 2026-07-06
Simulator Hand Reach Limit

whuang37:
Added new configuration opens for the simulator to limit the hand reach length and the radius of reach.

- New origin points for the hands to simulate shoulders
- checks if hands in range and stops movement if not
- Reach angle checks angle in direction of camera to limit angle of movement
  - defaults to 180 degrees for a realistic shoulder angle range

### DIFF
--- a/src/simulator/SimulatorOptions.ts
+++ b/src/simulator/SimulatorOptions.ts
@@ -92,6 +92,10 @@ export class SimulatorOptions {
     leftHandOrigin: {x: -0.2, y: -0.2, z: 0},
     rightHandOrigin: {x: 0.2, y: -0.2, z: 0},
   };
+  reachAngle = {
+    enabled: false,
+    angle: 180,
+  };
 
   constructor(options?: DeepReadonly<DeepPartial<SimulatorOptions>>) {
     deepMerge(this, options);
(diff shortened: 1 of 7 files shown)


dli7319 (approved):

src/simulator/SimulatorOptions.ts
```diff
@@ -86,6 +86,16 @@ export class SimulatorOptions {
   renderToRenderTexture = true;
   // Blending mode when rendering the virtual scene.
   blendingMode: 'normal' | 'screen' = 'normal';
+  reachDistance = {
```
ruofeidu, line 90: nit: add comment here.

src/simulator/SimulatorOptions.ts
```diff
@@ -86,6 +86,16 @@ export class SimulatorOptions {
...
+    enabled: false,
+    radius: 0.75,
+    leftHandOrigin: {x: -0.2, y: -0.2, z: 0},
+    rightHandOrigin: {x: 0.2, y: -0.2, z: 0},
+  };
+  reachAngle = {
```
ruofeidu, line 100: nit: add comment here.

src/simulator/SimulatorOptions.ts
```diff
@@ -86,6 +86,16 @@ export class SimulatorOptions {
...
+    enabled: false,
+    angle: 180, // Degrees around camera vector direction.
```
ruofeidu: nit: angleDegrees: 180

otherwise, people may misconcept with angleRadian

Three.js uses radians by default for all angles and rotations, such as the object.rotation property

please change to radian.

ruofeidu (changes requested):
LGTM, please add some comments for the new options

maybe consider update the xb-simulator skills as well.

ruofeidu (approved):

PR #429, opened 2026-07-07
Added NavMesh Support to Simulator

whuang37:
Simulator now can run on a navmesh to support bounded walking and realistic interactions in environments
- Added one new dependency [three-pathfinding](https://github.com/donmccurdy/three-pathfinding)
- created a navmesh for the simulator scene
- Added SimulatorNavMesh which sets up pathfinding and constrained movement throughout scenes on navmesh
- Added a demo for visualization

Leaving as draft since the navmesh isn't in the assets repo yet so the loading for that isn't done.

[demo](https://youtu.be/vwo7k1EaK7Q)

### DIFF
--- a/src/simulator/SimulatorNavMesh.ts
+++ b/src/simulator/SimulatorNavMesh.ts
@@ -9,10 +9,7 @@ const RANDOM_PATH_SAMPLE_ATTEMPTS = 8;
 
 type PathfindingConstructor = {
   new (): PathfindingType;
-  createZone: (
-    geometry: THREE.BufferGeometry,
-    tolerance?: number
-  ) => unknown;
+  createZone: (geometry: THREE.BufferGeometry, tolerance?: number) => unknown;
 };
 type PathfindingNode = ReturnType<PathfindingType['getClosestNode']>;
 type PathfindingZone = {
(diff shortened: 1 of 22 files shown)


src/simulator/SimulatorNavMesh.ts
```diff
@@ -0,0 +1,326 @@
+import * as THREE from 'three';
+import {GLTFLoader} from 'three/addons/loaders/GLTFLoader.js';
+import type {Pathfinding as PathfindingType} from 'three-pathfinding';
+
+import {SimulatorEnvironment, SimulatorOptions} from './SimulatorOptions';
+
+const DEFAULT_ZONE_ID = 'simulator';
+const RANDOM_PATH_SAMPLE_ATTEMPTS = 8;
+
+type PathfindingConstructor = typeof import('three-pathfinding').Pathfinding;
```
dli7319: I think we can avoid doing another import here somehow.
Maybe `typeof Pathfinding`?

dli7319 (approved):

src/simulator/SimulatorNavMesh.ts
whuang37: I just converted it to a static constructor definition. Could consider making the import non-lazy but i feel like its nice to keep it lazy.

ruofeidu (approved):

PR #401, opened 2026-06-21
Sensors Addon

whuang37:
Refactor all third person sensing into a new addon called ```sensors```

- ```SensorsManager``` manages all sensor lifecycles to enforce time alignment, dependencies, and caching
- Shifted all sensing to a unified frame-based caching which allows multiple scripts to query for the same sensing output
- ```Sensors<T>``` allow easy extension of new sensors
- Scheduling mode for sensors to allow for variable performance hits

New Sensors
- ```ScreenshotSensor``` captures screenshots of the raw world
- ```ScreenshotXR``` captures. screenshot of the world + XR
- ```ScreenshotSOM``` adds annotations over XR objects for referencing by AI
- ```VisibilitySensor``` traces what objects are in the user's viewport
- ```SemanticMapSensor```  formats visible sensors into plaintext descriptiosn
- ```DepthSensor``` depth maps
- ```TargetingSensors``` tracks what objects the hand is pointing at or colliding with
- ```ProprioceptionSensors``` tracks user transforms

GeminiManager Changes
- Movd the screenshot captures out of ```geminiManager``` and made it run from sensors instead.

New sample showing data at ```sample/sensors_debug```

PR #431, opened 2026-07-10
Context Module

whuang37:
Added the first context module for agents to better understand xrblocks. Adaptation from https://github.com/google/xrblocks/pull/401

- fixed a bug in ```screenshotSynthesizer``` where stacked async GPU readbacks could cause a memory leak when ```.onAfterRender()``` does multiple readbacks for the same pending screenshot
- added ```xb.context``` which now exposes the first context module ```xb.context.scene```
	- exposes three new data pieces:
		- ```xb.context.scene.tree``` which is a cut down semantic tree of objects designed for easy agent readability
		- ```xb.context.scene.visibleObjects``` which is a list of all objects within the view of the user including occlusion checked via raycasts
		- ```xb.context.setOfMark``` an overlayed image of the user's pov with marks that reference objects for agents to reference
		- Unique ids to reference objects in the scene
	- all 3 are dependent on each other so made into one detector with varying detection and enable settings
	- added a new demo for this data ```demo/context``` to show this data

Video demo [here](https://youtu.be/Zsn0Rumw4XU)


ruofeidu:
Thank you William for filling in the `xb.context` gap!!

src/context/shared/SemanticTypes.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export type VisibleObjectsContext = SemanticTree;
+
+export interface SetOfMark {
+  label: string;
+  nodeId: string;
+  role: string;
+  name: string;
+  screenCenter: Vec2Tuple;
```
ruofeidu: document (0, 0) as bottom-left somewhere? 
How was the Gemini object detector returning coordinates so far? like OpenCV or graphics?

ruofeidu (approved):

ruofeidu (approved):

src/context/shared/SemanticTypes.ts
whuang37: just changed this to origin from top left nromalized. Thats what object detection and faces in the repo currently do

dli7319 (approved):

PR #433, opened 2026-07-13
scene editor addon

Peter-Pater:
xrblocks scene editor addon, built in xrblocks/src/addons/editor/. Now provided as a simulator mode in xrblocks, allowing convenient scene editing and development, with familiar Blender/Unity style operations. This addon includes a full scene editor, with transform gizmo, object hierarchy, undo/redo, scene export/import, and object asset managing. 

## Features

**Scene population**
- Model picker (top-left): browse `Models/`, live 3D preview thumbnail, Spawn. The directory listing is polled every 3s, so files added/removed on disk while the app is running show up automatically without a reload.
- Every spawn is an independent instance — the same source file can be loaded any number of times.
- Duplicate / Delete, single- or multi-object, from the inspector panel or the `Delete`/`Backspace` key.

**Selection**
- Click to select; `Shift`+click to add/remove from a multi-selection (in the 3D scene or the hierarchy panel).
- `Ctrl+A` selects everything (except locked objects); `Esc` clears the selection.
- Per-object bounding-box highlight while selected.
- `F` (Frame Selected) repositions the camera to frame the current selection.

**Transform gizmo** (digits `1`–`4` switch tool mode: select / translate / rotate / scale)
- Axis-constrained translate (arrows + plane quads), arbitrary-axis rotate (rings), uniform + per-axis scale (center cube + axis handles).
- Multi-select transforms operate on the *whole group* around a shared pivot (the centroid of every selected object's own pivot) — rotate/scale a group and it behaves like a real rigid-body transform, not independent per-object spins. This reduces exactly to single-object behavior when only one thing is selected (by construction, not just by testing).
- World/Local space toggle (button in the inspector panel) — axis directions follow the primary selected object's own orientation in Local mode.
- Numeric inspector (top-right): typed position/rotation/scale fields. With a multi-selection, a field shows a value only if every selected object agrees on it, otherwise a blank "Mixed" placeholder; committing a value applies it to the whole selection at once.

**Hierarchy / outliner** (bottom-left, docked under the model picker)
- One row per instance, duplicate source files disambiguated with a trailing `#id`.
- Double-click a row's name to rename it (`customName`, shown everywhere in place of the auto-derived name).
- Per-row visibility toggle ("V") and lock toggle ("L"). Locked objects can't be selected at all (3D click or hierarchy click); hidden objects can't be 3D-clicked but stay selectable from the hierarchy so you can re-show them.
- Fixed-height list (~10 rows), scrolls beyond that.

**Undo/redo**
- `Ctrl+Z` / `Ctrl+Shift+Z`. Every gizmo drag, inspector edit, spawn, delete, and duplicate is undoable — including multi-object operations, which collapse into a single undo step via a batched command (`CommandHistory.pushBatch`).
- Rename, visibility, and lock toggles are the one exception: they're treated as organizational metadata, not scene content, and are **not** undoable (though they *are* correctly preserved across delete/duplicate undo).

**Scene persistence** (bottom-right panel)
- Export the current scene to a timestamped JSON file (browser download — this is a static file server with no backend, so getting a saved file into `Scenes/` for the Import picker requires moving it there manually once).
- Import replaces the whole scene from a picked `Scenes/*.json` file. Newer fields (`customName`/`visible`/`locked`) default gracefully when absent, so older scene files remain compatible.
- **This app auto-imports the first scene in `Scenes/` on startup**, if any exist (`main.js`, app-specific — not part of the addon; see [Model-viewer-specific extras](#model-viewer-specific-extras-not-part-of-the-addon)).
- Simulator-room occlusion toggle (off by default) — makes spawned objects actually hide behind simulator walls/furniture instead of drawing through them.

**Simulator "Editor" mode** — the editor's panels, gizmo, and highlights are visible/interactive only while the simulator's Interaction Mode (gear icon, bottom-left) is set to **Editor**. Switching to any other mode (User, Navigation, Hands, Pointer Lock) hides the whole editor UI immediately — non-destructively: your selection and undo history are preserved, so switching back to Editor restores exactly where you left off. This app boots directly into Editor mode so it's visible immediately, matching the pre-addon always-on behavior; other apps using the addon default to whichever mode they choose and opt in via the gear.

## Controls summary

| Input | Action |
|---|---|
| Click | Select (replaces selection) |
| `Shift`+click | Add/remove from selection |
| `Esc` | Clear selection |
| `Ctrl+A` | Select all (skips locked) |
| `F` | Frame selected |
| `1`/`2`/`3`/`4` | Tool mode: select / translate / rotate / scale |
| `Delete`/`Backspace` | Delete selection |
| `Ctrl+Z` / `Ctrl+Shift+Z` | Undo / redo |
| Double-click hierarchy row | Rename |

All of the above only fires while the simulator is in Editor mode (see above).

## Architecture

This folder itself is now minimal:

| File | Responsibility |
|---|---|
| `main.js` | Boots `xb`, adds a `SceneEditor` from the addon, plus two app-specific extras (see below) |
| `index.html` | Minimal shell — importmap + canvas + script tag. No panel markup or CSS: every panel injects its own. |

The actual editor lives in `xrblocks/src/addons/editor/` (TypeScript, built to `xrblocks/build/addons/editor/`):

| File | Responsibility |
|---|---|
| `SceneEditor.ts` | Public entry point / orchestrator. Constructs and wires everything below, owns the shared DOM root, owns Editor-mode + XR-session visibility gating. |
| `SceneManager.ts` | Instance registry: spawn/despawn, occlusion, auto-fit scale |
| `SelectionManager.ts` | Multi-select state, 3D-click hit-testing, highlights, keyboard shortcuts, the `editorActive` gate |
| `TransformGizmo.ts` | The gizmo itself — handle geometry, drag math, group-pivot transforms |
| `TransformInspectorPanel.ts` | Top-right numeric panel, Duplicate/Delete |
| `ModelPickerPanel.ts` | Top-left picker + live preview thumbnail |
| `HierarchyPanel.ts` | Bottom-left outliner: rename/visibility/lock |
| `ScenePanel.ts` | Bottom-right: occlusion toggle, export/import |
| `CommandHistory.ts` | Undo/redo stack, including batched multi-object commands |
| `styles.ts` | The shared stylesheet, injected into the page once |
| `dom.ts` | Tiny `el(tag, props, ...children)` helper for building DOM without a template-string/framework dependency |

Every panel builds and injects its own DOM + CSS at runtime (via `styles.ts`/`dom.ts`) rather than requiring the consuming app's `index.html` to pre-declare matching markup — a consuming app just does `xb.add(new SceneEditor({}))` and gets the whole editor, no HTML/CSS authoring required. This sidesteps the spatial UI toolkit's lack of dropdown/list/text-input widgets and its non-intuitive `fontSize` behavior, and gives real typed `<input type="number">` fields for free — same reasoning as the original 2D-overlay design, just self-contained now instead of app-provided.

## xrblocks changes

Everything below was added to the shared `xrblocks` framework (not app-local) to support this editor, ahead of a PR.

### Core framework (`xrblocks/src/`)

| Change | Purpose |
|---|---|
| `simulator/SimulatorOptions.ts` — added `SimulatorMode.EDITOR` enum value, extended `DEFAULT_MODE_TOGGLE_ORDER` | Makes "Editor" a real interaction mode, reachable both from the gear dropdown and the mode-cycle keyboard shortcut, alongside User/Navigation/Hands/Pointer Lock |
| `simulator/controlModes/SimulatorEditorMode.ts` (new) | `extends SimulatorUserMode` with no overrides — identical camera/click behavior to User mode. Exists purely as a distinct mode *key* so the editor addon can gate its own UI on `simulatorMode === EDITOR`, without changing how the user navigates or clicks |
| `simulator/SimulatorControls.ts` — registered `SimulatorEditorMode` in the mode lookup table | Wiring for the above |
| `addons/simulator/ui/SimulatorSettingsPanel.ts` — added "Editor" option | Surfaces the new mode in the gear/settings dropdown UI |
| `ui/interaction/ModelViewer.ts` — added a public `contentScene` getter | `gltfMesh` (the loaded GLTF) is `protected`; the editor needs to read/rotate the loaded content's root object from outside the class (for the gizmo, inspector, and scene export). The getter exposes just `gltfMesh.scene`, keeping the rest of `gltfMesh` encapsulated. |

### New addon (`xrblocks/src/addons/editor/`)

The entire editor described above, ported from what was originally app-local JavaScript in this folder — same behavior, now TypeScript, self-contained UI, and reusable by any app. See the [Architecture](#architecture) table for what each file does.

## Simulator "Editor" mode + XR safety net

Two independent layers of gating, both owned by `SceneEditor.update()`/`onXRSessionStarted()`/`onXRSessionEnded()`:

1. **Mode-based scoping**: the editor is visible/interactive only when `xb.core.simulatorRunning && xb.core.simulator.controls.simulatorMode === xb.SimulatorMode.EDITOR`. This is non-destructive — switching to another simulator mode just stops rendering/responding to input, it doesn't clear the selection or undo history, so returning to Editor mode restores exactly where you left off.
2. **Hard XR cutoff**: `xb.core.simulatorRunning` turns out to **never reset to `false`** once a real XR session starts (confirmed by reading `Core.ts` directly), so mode-based gating alone can't be trusted to mean "definitely not in XR." `SceneEditor` also overrides `onXRSessionStarted()`/`onXRSessionEnded()` as an independent, forceful cutoff — the instant a real headset session begins, the editor goes fully inert and the selection is cleared outright (unlike the ordinary mode-switch case, "resume editing later" isn't a meaningful concept once you've left the desktop/simulator context entirely).

Spawned scene content (`SceneManager`) is untouched by either layer — it's scene content, not editor chrome, and the whole point of the editor is to produce a scene that renders normally afterward, including in a real headset. Separately, `SelectionManager`/`TransformGizmo`'s 3D-click hit-testing has always gated on `event.target === xb.core.input.mouseController`, which a real XR controller never satisfies — pre-existing protection, not something this gating added.


whuang37:
@ruofeidu @dli7319 @Peter-Pater This is a super useful tool! Maybe instead of fitting into the direct sdk it can be a separate part of devtools? @Peter-Pater how important do you think the interaction flow is to switching to editor in simulator mode vs just fly around

dli7319 (approved):
Please fix the tsdoc warnings.

Peter-Pater:
> Please fix the tsdoc warnings.

Fixed :)

src/ui/interaction/ModelViewer.ts
```diff
@@ -90,6 +90,16 @@ export class ModelViewer extends Script implements Draggable {
...
   protected animationMixer?: THREE.AnimationMixer;
   protected gltfMesh?: GLTF;
+
+  /** The loaded GLTF content's root object, if a model has been loaded via
+   * loadGLTFModel() -- public so external code (e.g. a scene-editor addon
+   * applying its own rotation gizmo) can read/transform the loaded
+   * content directly, independent of the ModelViewer's own
+   * position/scale. Exposes just the scene root, not the full GLTF
+   * (animations/cameras/etc.), to keep the rest of gltfMesh encapsulated. */
+  get contentScene(): THREE.Object3D | undefined {
```
ruofeidu: discuss: is `modelScene` or `contentScene` a better name, as this is within `ModelViewer`.

ruofeidu (approved):

src/ui/interaction/ModelViewer.ts
Peter-Pater: Em, modelScene is indeed the more consistent naming. Changed.

ruofeidu (approved):

src/ui/interaction/ModelViewer.ts
dli7319: You can also just make `gltfMesh` public if you want. I actually don't see a particular need to keep it protected.

PR #438, opened 2026-07-15
Head Gesture Recognizer

whuang37:
Added head gesture recognizer for nodding and shaking your head.

- Added as ```options.enableHeadGestures()``` which you can subscribe to with gestures
- Follows same conventions as the existing gestureRecognition but just in input
- Changed ```Input``` so it initializes even when controllers are disabled and made controller enable not hinge all of input onto it.
- Head gesture algorithm as follows: Every frame ```HeadGestureRecognition``` records camera world space timestamp, position, orientation. Retains a 1.5 second window and evaluates every 16 ms. Discontinuities invalidate the window. Averages orientation across samples to get a resting orientation that follows the user. Then every orientation is evaluated relative to resting. To get shake and nod, we evaluate yaw and pitch accordingly to find an angular peak. Tuning parameters in ```HeuristicHeadGestureRecognizer``` to evaluate. Currently, must complete in 200-750 ms, reach .6 rad/s, limited off axis rotation, and follow a reasonable out and back pathway. Emits a completion when identified. Confidence computed via amplitude above threshold, gesture duration, closeness of return to baseline, off axis motion, and path efficiency.
- Set up in same way as the hand so we can load in ML models for this if we want.
- Documented

Tested on device and it works pretty well, better than in simulator for me but its an easy tune either way since its heuristic.

[Sim Demo](https://youtu.be/bYDlsnQP4eM)


dli7319:
Thank you! Should be good for glasses.

dli7319 (approved):

ruofeidu (approved):
Great work!
We are closer on the way productionizing [Sensible Agents](https://duruofei.com/projects/sensibleai/) :)

PR #444, opened 2026-07-20
keep the AgentHands controls closer to neutral gaze

salmanmkc:
the control panel from #416 sits pretty far below the initial view on Galaxy XR, so it needs a noticeable look down before you can use it.

this lifts the leash target from 0.55 m below gaze to 0.2 m and keeps it close at 0.9 m. the panel still follows smoothly and keeps its drag behavior, but should start inside a more comfortable part of the view without making the desktop rendering soft from pushing it farther away.

opening this as a draft until the placement gets a Galaxy XR check.

### DIFF
--- a/demos/agent_hands/main.js
+++ b/demos/agent_hands/main.js
@@ -61,6 +61,11 @@ const LEAN_CLAMP = 0.5; // max yaw lean toward a pointing target (rad)
 const LEAN_X_GAIN = 0.25; // pitch lean per unit target height
 const LEAN_X_CLAMP = 0.25; // max pitch lean (rad)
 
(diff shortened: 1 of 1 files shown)


salmanmkc:
@dli7319 could you try this placement on Galaxy XR and let me know whether the new height feels comfortable, or if you would still want it a little lower? I kept the distance close to the original so it stays sharp in the desktop simulator.

dli7319:
Looks good now! Thank you.
screenshot

salmanmkc:
> Looks good now! Thank you. screenshot

Awesome thanks for the quick test and review!

PR #453, opened 2026-07-21
GLTF Reuse Fix

whuang37:
Allowed each GLTF loader to use its own instance + ktx on load up. dedupes any paths and resets if paths already exist


src/utils/ModelLoader.ts
```diff
@@ -69,6 +58,19 @@ export class ModelLoader {
     this.manager = manager;
   }
 
+  private getGLTFLoader(renderer?: THREE.WebGLRenderer) {
```
ruofeidu, line 61: can we abstract and hide this implementation details in the SDK in future?

ruofeidu (approved):

src/utils/ModelLoader.ts
dli7319: This is private.

dli7319 (approved):

Issue #461, opened 2026-07-25
Fix Reticle Collision with Hand Mesh When Moving


whuang37:
Reticle collides with hand mesh and doesn't visualize properly when moving in simulation mode with raycasts. Causes the reticle to jitter in the simulation hand until stopped. Can't diagnose precisely when but will happen in some apps?

ruofeidu:
is that because `options.enableDepth()`? Then reticle collide with physical environment as well as hand mesh...

whuang37:
Oh could be since depth mesh is async and the hands are in the simulator depth mesh. potentially just adding a close distance guard for the reticle on depth mesh would fix?

ruofeidu:
I think so

PR #464, opened 2026-07-26
Sim Reticle Jitter Fix in Depth

whuang37:
Fix for the sim reticle jittering infront of the hand due to the depth mesh rendering async with the reticle intersection check https://github.com/google/xrblocks/issues/461.

Only could fix it fully by disabling reticle casting on the hands and then removing the hands from the depth creation and readding after. 

There's an argument to be had this isn't 1:1 with real depth sensing since we cull the hands from the sensing but the jitter was causing alot of issues with simulator + reticle apps for me. A distance check didnt really work since its dependent on framerate + user speed and itd have to be something along the lines of .4m+ to work and cause issues.

### DIFF
--- a/docs/docs/manual/Simulator.mdx
+++ b/docs/docs/manual/Simulator.mdx
@@ -184,6 +184,14 @@ The simulator also supplies simulated planes, meshes, object detections, a
 device-camera feed, and optional physics-backed environment objects when those
 subsystems are enabled.
 
+Simulated hands are excluded from generated depth data by default. This prevents
+hand rays from hitting the depth image of the same hand. Set the option to
+`false` when a test must include hand depth:
+
+```js
+options.simulator.excludeHandsFromDepth = false;
+```
+
 ## Hand reach and simulator physics
 
 Limit virtual hands to a shoulder-relative radius or forward cone:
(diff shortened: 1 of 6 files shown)


dli7319:
Fyi this often happens on device as well.
Since we need this for your use case, can we have a simulator option like `excludeHandsFromDepth`.
You can make it true by default.

whuang37:
> Fyi this often happens on device as well. Since we need this for your use case, can we have a simulator option like `excludeHandsFromDepth`. You can make it true by default.

Potentially could just change this to a hand distance check instead and cull all reticles within some distance to the hand. It wouldn't work 100% of the time but could catch both issues on device and sim. Could be breaking though for short distance reticles so idk if worth it? Just found in simulator this was a problem 100% of the time when doing any sort of strafing.

dli7319:
We used to have something like starting the raycast at 30 cm in front of the hand. Not sure if that's still there. I may have removed it by accident.

whuang37:
> We used to have something like starting the raycast at 30 cm in front of the hand. Not sure if that's still there. I may have removed it by accident.

dont think its there anymore. Is 30 cm too far away?

dli7319 (approved):

dli7319:
Yeah 30 cm is too far. Lets submit this for now. I'll investigate it on device when I have time.

ruofeidu (approved):

whuang37:
defunct after v0.20

PR #476, opened 2026-08-03
Interaction Module Implementation

whuang37:
## Description

**Not tested on device yet.**

This PR adds the new standalone interaction system on top of the lifecycle fixes.

The new module owns hit resolution, selection capture, gaze, touch, reticles, semantic controls, and object manipulation. The existing UI-dependent interaction path remains temporarily so this PR can be reviewed without also including the UI migration.

Stack dependency: `manip-lifecycle`

## Changes

- Add a standalone interaction module.
- Add hit target registration and hit resolution.
- Add selection capture across input sources.
- Add gaze dwell interaction.
- Add direct-touch interaction.
- Add reticle presentation and reticle configuration.
- Add semantic controls and interaction metadata.
- Add object manipulation for translation, rotation, and scaling.
- Add manipulation drivers and shared manipulation math.
- Suspend placement transform scripts while their objects are manipulated.
- Add the `Orbit` transform script.
- Add configurable FaceCamera and reticle rotation smoothing.
- Preserve the allocation-free FaceCamera implementation.
- Fold the formatting-only changes into the interaction implementation.

## Scope boundaries

This PR does not connect Core or Input to the new interaction module.

The following legacy files remain because the existing UI still depends on them:

- `UX`
- `DragManager`
- The old raycaster
- The old reticle
- The old UI-dependent interaction code

The runtime cutover and deletion of those files occur in the next PR.

### DIFF
--- a/src/interaction/reticle/Reticle.ts
+++ b/src/interaction/reticle/Reticle.ts
@@ -25,9 +25,6 @@ export class Reticle extends THREE.Mesh<
   /** Ensures the reticle is drawn on top of other transparent objects. */
   renderOrder = 1000;
 
-  /** The smoothing factor for rotational slerp interpolation. */
-  rotationSmoothing: number;
-
   /** The z-offset to prevent visual artifacts (z-fighting). */
   offset: number;
 
@@ -59,7 +56,7 @@ export class Reticle extends THREE.Mesh<
    * objects. Defaults to `false` to ensure it is always visible.
    */
   constructor(
-    rotationSmoothing = 0.8,
+    public rotationSmoothing = 0.8,
     offset = 0.001,
     size = 0.019,
     depthTest = false
--- a/src/interaction/reticle/Reticle.ts
+++ b/src/interaction/reticle/Reticle.ts
@@ -1,0 +1,201 @@
+import * as THREE from 'three';
+
+import {lerp} from '../../utils/utils';
+import {ReticleShader} from './ReticleShader';
+
+const HOVER_RING_BRIGHTNESS = 0.4;
+const HOVER_RING_OPACITY = 0.7;
+
+/**
+ * A 3D visual marker used to indicate a user's aim or interaction
+ * point in an XR scene. It orients itself to surfaces it intersects with and
+ * provides visual feedback for states like "pressed".
+ */
+export class Reticle extends THREE.Mesh<
+  THREE.BufferGeometry,
+  THREE.ShaderMaterial
+> {
+  /** Text description of the PanelMesh */
+  name = 'Reticle';
+  editorIcon = 'target';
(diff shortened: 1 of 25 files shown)


src/interaction/reticle/Reticle.ts
```diff
@@ -0,0 +1,201 @@
...
+      new THREE.ShaderMaterial({
+        uniforms: THREE.UniformsUtils.clone(ReticleShader.uniforms),
+        vertexShader: ReticleShader.vertexShader,
+        fragmentShader: ReticleShader.fragmentShader,
+        depthTest: depthTest,
+        transparent: true,
+      })
+    );
+
+    this.rotationSmoothing = rotationSmoothing;
```
dli7319, line 78: Nit: We can declare this as `public` in the constructor params.

dli7319 (commented):
This doesn't integrate it right?
The existing reticle is still there.

dli7319:
>Allows independent objects to be translated concurrently when their sessions do not conflict.

Will two hand pinch not scale anymore? Or is this only when the underlying object isn't scalable or has a different scale gesture?

whuang37:
> >Allows independent objects to be translated concurrently when their sessions do not conflict.
> 
> Will two hand pinch not scale anymore? Or is this only when the underlying object isn't scalable or has a different scale gesture?

when scales enabled itll take over both hands to scale like before but when scale isnt enabled itll do dual selection

whuang37:
> This doesn't integrate it right?
> The existing reticle is still there.

yea this is just the implementation. integration requires full breaking with existing adk

dli7319:
From AI reviewer:
> 🟡 FaceCamera.ts (update()) and FaceCameraMath.ts (faceCameraQuaternion()) instantiate new THREE.Vector3, THREE.Quaternion, and THREE.Matrix4 instances on every frame. Reuse static scratch instances (matching FollowHead.ts and FollowObject.ts) to eliminate GC pauses during WebXR render loops.

src/singletons.ts
```diff
@@ -101,26 +101,6 @@ export function init(options: Options = new Options()) {
   return core.init(options);
 }
 
-/**
- * A shortcut for `core.scriptsManager.initScript()`. Manually initializes a
- * script and its dependencies.
- * @param script - The script to initialize.
- * @see {@link ScriptsManager.initScript}
- */
-export function initScript(script: Script) {
```
dli7319, line 109: Can we leave this in?

dli7319 (approved):

src/singletons.ts
whuang37: yea this was added back in the previous stack. let me rebase back in the dev branch and fix this

PR #477, opened 2026-08-03
Runtime Integration on Refactor

whuang37:
## Description

**Not tested on device yet.**

This PR connects the runtime to the new interaction system and replaces the legacy UI and UIBlocks implementations with the built-in spatial UI kit.

It performs the final interaction cutover, migrates consumers, and removes the old UI-dependent interaction path without compatibility aliases or transitional shims.

Stack dependency: `manip-02-interaction`

### Runtime interaction cutover

- Connect Core, Input, User, and Script callbacks to the new interaction module.
- Route hover, selection, touch, gaze, and manipulation through the new interaction service.
- Update simulator input wiring where required by the interaction cutover.
- Replace old reticle and ModelViewer interaction dependencies.

### Built-in spatial UI

- Add the built-in UI kit and renderer.
- Add UI themes, elements, controls, cards, overlays, sliders, text, images, and icons.
- Add the migrated ModelViewer implementation.
- Move reusable UI primitives from UIBlocks into the built-in UI module.

### Consumer migrations

- Migrate the editor.
- Migrate the glasses addon.
- Include the glasses `TextWithEmoji` fix.
- Migrate testing and remote-control consumers.
- Migrate Netblocks samples and consumers.
- Update the UI template.

### Legacy removal

- Remove the old UI implementation.
- Remove the UIBlocks addon.
- Remove `UX` and `DragManager`.
- Remove the old raycaster and reticle.
- Remove the old UI-dependent interaction files.
- Remove obsolete exports, options, package entries, and Rollup entries.
- Remove obsolete UIBlocks documentation and skill links.
- Update documentation, templates, and skills for the new UI interface.

No aliases or compatibility shims are included.

## Public API changes

- Export the new interaction event model.
- Export object interaction metadata.
- Export the built-in spatial UI interface.
- Remove legacy UI and UIBlocks exports.
- Remove `UX`, `DragManager`, and the old reticle interfaces.

## Scope boundaries

Simulator lifecycle cleanup, depth, occlusion, device-camera disposal, physics disposal, and semantic-context changes remain in the next PR.

Examples and demos remain in the final PR.


dli7319:
So XR Blocks will now require @pmndrs/uikit?

whuang37:
> So XR Blocks will now require @pmndrs/uikit?

this is the one thing im currently still playing with. 

as of this moment **no**. right now i effectively rewrite all the signatures from uikit and call uikit via composition. then on build we turn from xrblocks.js to a directory where the uikitrenderer backend is in a chunk. only when ui is added will this chunk be needed which will require uikit. otherwise uikit will never be required since its not exposed by inheritance.

we talked about it and requiring uikit could be ok since ui is so fundamental or keeping it in addons but i went with this for now since it reduces the size of everything and since manually defining ui elements lets me clean up the overall scene graph, add custom ui like sliders, and use our own raycasting over uikits anyways.

if we like this approach i could see an argument for it for the ai features as well later on

ruofeidu:
At this point I'm fine introducing uikit as long as it doesn't significantly increase the complexity and loading time of a simple XR drawing app :)

ruofeidu (approved):

PR #482, opened 2026-08-04
ci: fix zizmor security audit findings in GitHub Actions workflows

dli7319:
- Add explicit least-privilege permissions blocks to workflows
- Pin third-party and official GitHub Actions to SHA commit hashes
- Set persist-credentials: false on checkout steps
- Disable default caching in publish workflow to prevent cache poisoning

TAG=agy
CONV=a7043de3-4652-4ae1-b37e-aa75ef0ef8c2

### DIFF
--- a/.github/workflows/build-branch.yml
+++ b/.github/workflows/build-branch.yml
@@ -22,17 +22,19 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: ⬇️ Checkout Repo
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
 
       - name: ⚙️ Setup Node.js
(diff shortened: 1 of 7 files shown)


.github/workflows/build-branch.yml
```diff
@@ -45,7 +47,7 @@ jobs:
         run: npm run build
 
       - name: 🚀 Deploy to build branch
-        uses: peaceiris/actions-gh-pages@v4
+        uses: peaceiris/actions-gh-pages@329bcc8f12caed2cefe5a5b80781499a6f3b361b # v4
```
salmanmkc:
```suggestion
        uses: peaceiris/actions-gh-pages@84c30a85c19949d7eee79c4ff27748b70285e453 # v4.1.0
```

salmanmkc (changes requested):
think you need to change this one

ruofeidu (approved):

PR #483, opened 2026-08-04
fix(demos): run SAM fallback in isolated workers

salmanmkc:
Detect could pick a GPU with a 128 MiB binding limit even though SlimSAM needs 384 MiB for fp16 attention or 768 MiB for fp32. forcing WASM on the page instead froze rendering. checks both binding and buffer limits on the adapter passed to ONNX Runtime, plus `shader-f16` for fp16.

each candidate now runs in a fresh module worker, including WASM. worker isolation makes fallback recover from rejected initialization in the pinned Transformers.js `3.0.0` runtime, rather than retrying its poisoned module-level promise. verified that promise behavior by executing the published CDN function with a mocked session factory: the old runtime never reaches the WASM factory after the first rejection; a fresh runtime does.

only keep a candidate after its first snapshot encode succeeds. later inference failures discard it and retry the same snapshot on the next candidate. processor loads before the model, abandoned models are disposed, and both returned embedding tensors are reused for box prompts instead of accidentally re-encoding. `?mask=segmenter` stays an explicit choice.

mocked lifecycle tests cover capacity boundaries, rejected initialization, failed encoding, disposal, retry and worker failure. no Android XR or Quest hardware verification yet.


salmanmkc:
@dli7319 does this fix the GPU errors in the console for you?

dli7319:
The WebGPU errors are gone but the whole app completely freezes up now during detection when running on device 😅.

Screenshot 2026-08-04 at 10 57 05 AM

salmanmkc:
> The WebGPU errors are gone but the whole app completely freezes up now during detection when running on device 😅.
> 
> Screenshot 2026-08-04 at 10 57 05 AM

ya hard to test since my device doesn't hit this memory limit, seeing as it works but is just slower, I believe it has it's own fallback and forcing CPU blocks the thread, so I think best thing to do is leave the error message for now since it works

dli7319:
I tested this PR on Quest 3 and the same WebGPU errors appear there

Screenshot 2026-08-09 at 5 30 44 PM

PR #485, opened 2026-08-04
Lifecycle Updates for Core + Scripts

whuang37:
## Description

**Not tested on device yet.**

This PR improves Core and Script lifecycle reliability. It prevents initialization and disposal races, makes cleanup deterministic, and keeps script lifecycle management internal to Core.

This is the first PR in the stack and targets `dev`, which includes the merged placement work from #475.

## Changes

- Add explicit Core lifecycle states for initialization, running, disposal, and failure cleanup.
- Make `Core.dispose()` asynchronous and idempotent.
- Prevent initialization and disposal from running concurrently.
- Clean up partially initialized resources after an initialization failure.
- Improve tracking of pending, active, failed, removed, and disposed scripts.
- Centralize script callback dispatch and error reporting.
- Add script-error listeners through `Core.onScriptError()`.
- Remove repeated lifecycle allocations and next-frame synchronization delays.
- Make `ScriptsManager` internal to Core.
- Migrate Netblocks, Remote Control, and TestRunner away from direct `ScriptsManager` access.
- Clean up Core-owned resources:
  - XR session listeners
  - XR buttons
  - Render loops
  - Resize listeners
  - Renderer containers
  - WebXR session management
  - Effects and physics timers

Interaction behavior and the existing UI system are unchanged in this PR.

## Public API changes

- `Core.dispose()` now returns a Promise and is idempotent.
- Add the script-error listener interface.
- Remove the public `ScriptsManager` export.
- Remove the public `initScript` and `uninitScript` shortcuts.

### DIFF
--- a/src/addons/testing/example.test.ts
+++ b/src/addons/testing/example.test.ts
@@ -1,7 +1,7 @@
 import {describe, it, expect} from 'vitest';
 import {TestRunner} from './TestRunner';
 import * as THREE from 'three';
-import {Script, type SelectEvent, TextButton, Options, core} from 'xrblocks';
+import {Core, Script, type SelectEvent, TextButton, Options} from 'xrblocks';
 
 class SimpleRotationScript extends Script {
   speed = 1.0;
@@ -125,7 +125,7 @@ describe('TestRunner functional examples', () => {
     button.onTriggered = () => {
       game.spawnItem();
     };
-    button.position.set(0, core.user.height, -core.user.height);
+    button.position.set(0, 1.6, -1.6);
 
     const runner = await TestRunner.create({
       scripts: [game, button],
@@ -147,11 +147,15 @@ describe('TestRunner functional examples', () => {
 
   it('should run Example 5: End-to-End Heuristic Gesture Recognition', async () => {
     class TestGestureScript extends Script {
+      static dependencies = {core: Core};
+
       pinchDetected = false;
+      private core?: Core;
       private _onGestureStart?: (event: Event) => void;
       private _onGestureEnd?: (event: Event) => void;
 
-      override init() {
+      override init({core}: {core: Core}) {
+        this.core = core;
         const gestures = core.gestureRecognition;
         if (!gestures) return;
 
@@ -173,7 +177,7 @@ describe('TestRunner functional examples', () => {
       }
 
       override dispose() {
-        const gestures = core.gestureRecognition;
+        const gestures = this.core?.gestureRecognition;
         if (gestures) {
           if (this._onGestureStart) {
             gestures.removeEventListener('gesturestart', this._onGestureStart);
--- a/src/core/ScriptHooks.ts
+++ b/src/core/ScriptHooks.ts
@@ -1,0 +1,15 @@
+const defaultScriptMethods = new WeakSet<object>();
+
+/** Records the no-op methods supplied by one ScriptMixin base class. */
+export function markDefaultScriptMethods(prototype: object): void {
+  for (const name of Object.getOwnPropertyNames(prototype)) {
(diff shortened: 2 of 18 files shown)


dli7319:
IMO `initScript` is useful if you want to add something at runtime and immediately call a helper method without waiting around until the next frame.

dli7319:
Why are we using `onScriptError` instead of using threejs events?
This seems like we're adding a boutique event system in Core.

whuang37:
> Why are we using `onScriptError` instead of using threejs events? This seems like we're adding a boutique event system in Core.

it was originally a formalization for the testing addon since that does a capture of script errors for readback later. not super necessary here

whuang37:
> IMO `initScript` is useful if you want to add something at runtime and immediately call a helper method without waiting around until the next frame.

Didn't think about this makes sense!. let me move this back in just as a side option

src/core/ScriptHooks.ts
```diff
@@ -0,0 +1,15 @@
...
+export function markDefaultScriptMethods(prototype: object): void {
+  for (const name of Object.getOwnPropertyNames(prototype)) {
+    if (name === 'constructor') continue;
+    const method = Reflect.get(prototype, name);
+    if (typeof method === 'function') defaultScriptMethods.add(method);
+  }
+}
+
+/** Returns whether a Script method is supplied by ScriptMixin unchanged. */
+export function isDefaultScriptMethod(method: unknown): boolean {
```
dli7319, line 13: Is this used anywhere?
I'm not sure I understand the use case here.

dli7319 (approved):
Afaik, we're never actually calling core.dispose anywhere right now, but thanks for filling it out.

src/core/ScriptHooks.ts
whuang37: This is mainly used as an indexing. ScriptManager.indexScript() (417) on ScriptMixin creation. its basically just a check to lower the amount of hook calls we do by only indexing ones that have been overridden

src/addons/testing/example.test.ts
```diff
@@ -125,7 +125,7 @@ describe('TestRunner functional examples', () => {
     button.onTriggered = () => {
       game.spawnItem();
     };
-    button.position.set(0, core.user.height, -core.user.height);
+    button.position.set(0, 1.6, -1.6);
```
ruofeidu, line 128: idea: not a bug, but initially I'd prefer object / panels to be spawned dynamic to the front of the user with minimum tool calls.

user.height should be updated dynamically in core with average filter in the first 5 seconds.
user.objectDistance should be something like 1.5

src/core/ScriptHooks.ts
dli7319: Does this actually improve performance?

If a script does
```ts
private update = () => {...}
```
Then it won't get callbacks right?

ruofeidu (approved):

src/core/ScriptHooks.ts
whuang37: i think in prototype situations its negligible but its mainly for situations where theres a ton of scripts or steps like with onPhysicsStep. I originally added it cause i realized there would be a ton of onSelect, onSqueeze, onHover, etc that could be added with the updated interaction module and then they'd each add their own empty no-op call.

I think the major trade off is indexing prevents dynamic assignment of stuff like script.update.

dli7319: Ack

src/addons/testing/example.test.ts
whuang37: interesting. i think this could go hand in hand with the new concept of a placement script or like a spawn manager or something similar. followHead right now does something similar but just constantly updates

PR #488, opened 2026-08-05
Add XREAL Aura 

salmanmkc:
add xreal aura as it's supported, might help with SEO?


salmanmkc:
forced push to remove the typo

dli7319:
Great idea

PR #489, opened 2026-08-05
perf(world/humans): run MediaPipe PoseLandmarker in a web worker

### LINKED ISSUE #362: Move mediapipe vision to a web worker
After IIFE bundles are uploaded to npm, move Pose landmark logic to a webworker to not block the main thread.

salmanmkc:
closes #362.

pose detection ran `PoseLandmarker.detect()` on the main thread. it's synchronous and takes tens of ms, so every detection pass dropped frames. same thing faces had before #366.

moves it into a worker. rather than copy the plumbing out of the faces backend, pulled it into a shared client parameterised by task name so faces moves onto it too, one copy of the blob-url spawn / id correlation / imagebitmap transfer instead of two. only the landmark to world-space projection stays on the render thread, it needs the live depth mesh and camera matrices.

the trade: workers have no dom canvas so mediapipe won't spin up a gpu surface, worker inference is cpu where main-thread was gpu. for XR a dropped frame is worse than slightly slower inference (same call #366 made for faces) but it is a trade, so `useWorker: false` keeps gpu on the main thread. it also falls back on its own when there's no `Worker` or the worker fails to boot, before this a worker failure would have taken the backend down entirely.

`processPoseLandmarkerResult` comes out as a free function on the way, mirroring `processFaceLandmarkerResult`, so the projection is testable without standing up the backend.


dli7319 (approved):
Tested on device and the FPS is way smoother. Thank you!

PR #491, opened 2026-08-06
Disable matchDepthView by default and update occlusion

dli7319:
## Description

This CL updates occlusion to support matchDepthView=false and sets it to false by default now that occlusion and depth mesh no longer require matchDepthView.

When matchDepthView is false, the WebXR depth sensor may have a different pose (view matrix) and projection matrix than the rendering camera.

This change updates the occlusion pipeline to support matchDepthView=false:
1. Both Z-depth (vVirtualDepth) and texture UV coordinates (vTexCoord) are computed in the depth camera's clip and projection space (uDepthViewMatrix and uDepthProjectionMatrix).
2. Out-of-bounds fragments (outside [0, 1] UV or behind the depth camera) are safely treated as unoccluded (1.0), preventing edge clamping artifacts for objects outside the depth sensor's FOV.
3. Unused rightDepthTexture setDepthTexture call is removed from Depth.renderOcclusionPass() since occlusion currently renders a single mono occlusion map.

## Type of Change

## Checklist

- [x] **Tested in simulator & device**: Ran occlusion scene on galaxy XR. Verified occlusion lines up with VST.

### DIFF
--- a/src/depth/occlusion/OcclusionPass.test.ts
+++ b/src/depth/occlusion/OcclusionPass.test.ts
@@ -1,0 +1,157 @@
+import {describe, it, expect} from 'vitest';
+import * as THREE from 'three';
+
+import {OcclusionMapMeshMaterial} from './OcclusionMapMeshMaterial';
+import {OcclusionPass} from './OcclusionPass';
+
+describe('OcclusionMapMeshMaterial', () => {
+  it('declares uDepthViewMatrix and uDepthProjectionMatrix in uniforms and initializes with camera matrices', () => {
+    const camera = new THREE.PerspectiveCamera();
+    camera.position.set(1, 2, 3);
+    camera.updateMatrixWorld(true);
+
+    const material = new OcclusionMapMeshMaterial(camera, true);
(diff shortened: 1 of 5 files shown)


src/depth/occlusion/OcclusionPass.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    const camera = new THREE.PerspectiveCamera();
+    const occlusionPass = new OcclusionPass(scene, camera);
+
+    const texture = new THREE.Texture();
+    const depthViewMatrix = new THREE.Matrix4().makeTranslation(10, 20, 30);
+    const depthProjectionMatrix = new THREE.Matrix4().makeScale(2, 2, 2);
+
+    occlusionPass.setDepthTexture(
+      texture,
+      0.01,
```
ruofeidu: nit: /*=*/

ruofeidu (approved):

src/depth/occlusion/OcclusionPass.test.ts
dli7319: Added parameters names.

PR #494, opened 2026-08-06
feat(world/anchors): persistent spatial anchors

salmanmkc:
## Description

Would like to record an on device demo but here's a 'simulation' of it on a laptop:

https://github.com/user-attachments/assets/af8d0563-7748-4b9b-a8f0-806051568d8b

`src/world/anchors/`. content you place in a room comes back in the same physical spot after you leave and re-enter the session.

been wanting to build this for xrblocks for a while. i did some research with azure spatial anchors about 5 years ago and persistence was something I used to build my shared supermarket experience, so getting it onto the open web stack was something i was pretty excited to try. kept stalling because i was abroad without my headset, and chrome on a phone implements the anchors module but not the persistence half of it, so all i could prove was that anchors got created. finally got back to my quest 3 and could actually test the thing properly.

anchors matter because a saved position drifts. the platform keeps refining its map of the room, so a coordinate you wrote down last week points somewhere slightly wrong today. an anchor gets corrected as that understanding improves, which is why `agenthands` currently saving raw world coordinates doesn't survive a session.

`xb.core.world.anchors` gives you `create` / `persist` / `restoreAll` / `getPose` / `delete`. every anchor api in webxr is optional and they're independent of each other, so capability is probed rather than assumed and reports one of `persistent`, `session-only`, `simulated` or `unsupported`. nothing throws on a platform without support, creation just returns null.

worth knowing: only the headset runtimes implement persistence. per mdn, `requestPersistentHandle` / `restorePersistentAnchor` / `deletePersistentAnchor` / `persistentAnchors` are recorded against oculus browser 31.2 and nothing else, and chromium's own `xr_anchor.idl` declares just `anchorSpace` and `delete()`. so on a phone you get `session-only` and the demo says so instead of silently failing. there's no separate `persistent-anchors` feature descriptor to request, the module defines exactly one, `anchors`.

what actually gets stored is worth being explicit about, since it differs by platform. on a headset a saved record is `{uuid, label, createdAt}` and nothing else: the uuid is the opaque handle from `requestPersistentHandle`, and there are no coordinates in it at all. the spatial part lives in the platform's own map, which is why a restore can come back correctly placed after the room has been re-scanned, and why it can honestly report not-found somewhere else. records go in `localStorage` under a key scoped to the page's directory so two demos on one origin don't rebuild each other's content, and keyed on the directory rather than the raw path because a demo reached as `/demos/anchors/` and as `/demos/anchors/index.html` would otherwise get two separate stores, which on a headset looks exactly like anchors going missing, and the store is behind an `AnchorStore` interface so an app that wants handles on a server can swap it.

the simulator record is the one that carries a pose, because nothing else is going to remember it. that's the tell that it isn't real anchoring, along with the `sim-` prefix on its handles.

desktop has no tracking system to anchor against at all, and there's no `XRFrame` outside an immersive session, so there's an opt-in `simulatorFallback` that holds poses locally. it proves app wiring only, never re-localisation, and the status line always says which backing is in use so that distinction can't be missed.

three demos: `demos/anchors/` drops markers, `demos/anchors_notes/` pins a typed note where you're looking, with the note text as the anchor label so what persists is real content, and `demos/anchors_gallery/` leaves shapes around the room through the `AnchoredObjects` helper, which owns the anchor-to-object map and copies poses each frame so an app doesn't have to. all three use a spatial panel rather than dom buttons since xrblocks doesn't request `dom-overlay`, which means the page is gone the moment you enter xr.

also exports `src/utils/ThreeDisposal` from the barrel. six subsystems already use it but it wasn't reachable from an app, so the demos had hand-rolled the same traversal.

to try it, `npm run dev` and open `demos/anchors_home/` on a headset, which lists the three and links between them the same way the netblocks samples index does. each one has an in-scene panel, so the controls are still there once you're in xr, and the status line on the panel tells you which backing you got. you want it to say "real anchors, saved across sessions" before any of this means anything, on a phone it'll say the platform can't save them and on desktop it'll say simulated.

`demos/anchors/` is the one to start with. **Drop** puts a marker 80cm in front of wherever you're facing. it doesn't hit-test against the room, so it floats there rather than landing on a surface, which is worth knowing when you go looking for it again. drop two or three next to something you'll recognise, then leave the session and come back in. they should be exactly where you left them. walk to another room and re-enter and they should report as not found rather than turning up in the wrong place. **Clear** removes them and forgets the saved handles, and it'll tell you whether the device actually released them.

`demos/anchors_notes/` is the same idea with real content. type on the spatial keyboard, hit **Pin**, and the note is anchored where you were looking with the text as its label. useful for leaving yourself something on the kettle or a door.

`demos/anchors_gallery/` is the same idea one layer up. pieces are placed at the angle you were looking and come back at that angle, so it's the whole pose being restored rather than a point, and `AnchoredObjects` is doing that per-frame work instead of the app. shapes that show which way they face, for that reason.

colour means the same thing in all three: something you just placed this session is purple in the marker and notes demos and orange in the gallery, and anything green came back out of storage. so after a reload everything should be green, and if you then place one more it stands out as the only one that isn't. that's the quickest way to tell a genuine restore from content that was simply never cleared.

all three keep their anchors in separate stores, so you can go between them without one rebuilding another's content.

in an app it's three lines:

```js
const options = new xb.Options();
options.world.enableAnchorPersistence();

// in your Script, once you have somewhere to pin
const tracked = await xb.core.world.anchors.create(pose, 'my label');
await xb.core.world.anchors.persist(tracked.id);

// next session
for (const {record, status, anchor} of await xb.core.world.anchors.restoreAll()) {
  if (status === 'restored') placeContentAt(anchor.id, record.label);
}
```

then read `getPose(id)` each frame and copy it onto your object, or hand that job to `AnchoredObjects` and skip the bookkeeping entirely, which is what the gallery demo does.

testing this on a quest turned up something worth writing down, since nothing in the spec prepares you for it. `persist()` can be refused, and it happens sooner than you'd think:

```
InvalidStateError: Failed to execute 'requestPersistentHandle' on 'XRAnchor':
Maximum number of anchors reached!
```

the budget belongs to the browser and is shared by every site in it, not per page, so another origin can be holding the slot you want. it's also small: mine refused the sixth anchor. the spec lets a user agent cap it, and lets it evict an entry to make room when every origin's anchors together hit the system maximum ([immersive-web/anchors#79](https://github.com/immersive-web/anchors/issues/79)), but nothing appears to do that evicting yet, so in practice once it's full it stays full until something explicitly releases handles. worth knowing that the platform can refuse a new anchor while reporting no handles at all for your origin, which leaves the page with nothing it can name and nothing it can free.

that makes a refused save an ordinary state rather than an edge case, and there's no way to check remaining capacity first. so the demos show the platform's own words instead of a generic failure, which is the difference between "it didn't work" and knowing exactly why. i had this blaming storage at first and lost an afternoon to it.

it also means a handle you lose track of is worse than useless: it holds its slot and nothing can name it. `forgetAll()` only reaches handles it recorded, so `releaseAllPlatformHandles()` reads the platform's own list instead, and each demo has a **Release** next to **Clear** for exactly that. it's origin wide, so it's an explicit action rather than anything automatic.

two quirks worth knowing if you go near `persistentAnchors`: it contains a blank entry even on an origin that has never saved anything, and it doesn't shrink as handles are released during a session, so reading it back to confirm a release tells you nothing. both are handled, and both are in the skill doc along with what to do when the budget is full.

open question on placement: i put this in `src/world/anchors/` next to `planes/` and `mesh/`, since those are also thin wrappers over a native webxr world-sensing feature and it ends up as one line in `Core.ts` alongside them. but the usual route here is demo first and promote later, which is what `objects_3d` did before it became the `objects3d` addon, and a demo can already push `anchors` onto `options.webxrOptionalFeatures` itself, so nothing forces it to live in the sdk. happy to pull the subsystem out and land the demos alone first if you'd rather see it proven that way, the demos are the part worth playing with anyway.

sat on pushing more while #417 was open, didn't want to put too much in front of you to review at once.

## Type of Change

- [ ] Bug fix
- [x] New demo or sample
- [ ] Documentation update

- **Simulator Recording**: top of description
- **Device Recording**: none yet, would like to record a proper one on the quest 3 at some point. the bit worth filming is walking out of the room and back in, and seeing an anchor from the next room through the wall.

## Checklist

- [x] **Tested in simulator & device**: verified on meta quest 3. dropped 2 markers in one room, left the session, came back and they were in the same places. walked to another room placed another anchor and they correctly reported as not found rather than appearing in the wrong spot, went back to the original room, it found all the original anchors (2) and the third one was visible through the wall from the room (second room) it was anchored in. desktop simulator path covered by the fallback above.
- [x] **Large Assets ($\ge$ 1MB)**: none added.
- [x] **SDK Dynamic Dependencies**: no new dependencies. the notes demo uses the existing `virtualkeyboard` addon.
- [x] **Security**: no keys or secrets. handles are opaque platform strings kept in local storage under a page scoped key.


salmanmkc:
above commit is since #495 merged in

it would be interesting to know the limits of android XR since meta quest has a low limit for anchors, at only 8

dli7319:
>the budget belongs to the browser and is shared by every site in it, not per page, so another origin can be holding the slot you want. it's also small: mine refused the sixth anchor. the spec sets no limit at all and browsers aren't currently allowed to evict old anchors to make room (https://github.com/immersive-web/anchors/issues/79), so once it's full it stays full until something explicitly releases handles.

Wow so a malicious page could essentially render anchors unusable by all other pages

salmanmkc:
> > the budget belongs to the browser and is shared by every site in it, not per page, so another origin can be holding the slot you want. it's also small: mine refused the sixth anchor. the spec sets no limit at all and browsers aren't currently allowed to evict old anchors to make room ([immersive-web/anchors#79](https://github.com/immersive-web/anchors/issues/79)), so once it's full it stays full until something explicitly releases handles.
> 
> Wow so a malicious page could essentially render anchors unusable by all other pages

Yeah, in building this I actually made a page as part of debugging that was meant to create an anchor and then delete it immediately after, I couldn't get that fully working since I had too many anchors already.

My undestanding is, that what makes it stick is that the page on the receiving end has nothing it can do about it. The list of persistent anchors you get off the session is scoped to your origin, but the budget is device wide, so when the slots are held elsewhere your list comes back empty and there's nothing to delete. I hit that state on my own origin a few times while testing, with `requestPersistentHandle` failing on "Maximum number of anchors reached!" at the same time as `session.persistentAnchors` returned nothing at all.

I should also correct what I wrote above: browsers are allowed to evict now. I did try but maybe it's some issue on the quest. The spec was updated after that issue was filed and says the UA may free up an entry once the maximum is reached across all origins. It doesn't look like anything implements it yet, so it's an implementation gap rather than something the spec forbids.

salmanmkc:
Tried to share some of my findings there too: https://github.com/immersive-web/anchors/issues/79#issuecomment-5207832958 

Hopefully there will be some better fix for it in the future on the quest

dli7319:
vlcsnap-2026-08-08-09h13m06s999

It's not working on my Galaxy XR 😬. We'll need to check with Alex to see if it's implemented or not in Chrome.

salmanmkc:
Had a dig into this. Persistence is specced, but yeah seems to not be in Chromium, seems like Meta is the only one who has implemented it yet, which is strange since on openxr it was made with Google, msft, meta and bytedance.

The web side is in the anchors module, `XRSession` gets `persistentAnchors`, `restorePersistentAnchor()` and `deletePersistentAnchor()`: https://immersive-web.github.io/anchors/#anchor-creation

And there's a ratified OpenXR extension behind it, `XR_EXT_spatial_persistence`, co-authored by Google (Nihav Jain, Jared Finder) with Meta, Microsoft and ByteDance. Documented for Android XR here: https://developer.android.com/develop/xr/openxr/extensions/XR_EXT_spatial_persistence

But Chromium doesn't ask unfortunately for it. `GetRequestedExtensions()` in `device/vr/openxr/openxr_spatial_framework_manager.cc` lists six extensions and persistence isn't one of them:

```cpp
XR_EXT_FUTURE_EXTENSION_NAME,
XR_EXT_SPATIAL_ENTITY_EXTENSION_NAME,
XR_EXT_SPATIAL_ANCHOR_EXTENSION_NAME,
XR_EXT_SPATIAL_PLANE_TRACKING_EXTENSION_NAME,
XR_ANDROID_SPATIAL_DISCOVERY_RAYCAST_EXTENSION_NAME,
XR_ANDROID_SPATIAL_ENTITY_BOUND_ANCHOR_EXTENSION_NAME,
```

https://source.chromium.org/chromium/chromium/src/+/main:device/vr/public/mojom/vr_service.mojom;l=1156

There's no plumbing for it further down either, `device/vr/public/mojom/vr_service.mojom` only has `DetachAnchor`, nothing to persist or restore. So the runtime support looks like it's there on Android XR and it's the Chromium side that's missing, maybe Alex knows about that bit. My guess is it needs a permission story first, since a UUID pinned to a physical spot that survives across sessions is a tracking surface, and the OpenXR extension itself gates it behind scopes.

Good news is I guess partially working and gracefully handling the error

dli7319:
It's working well on quest. The anchor spheres stay in the same place across recenter.

salmanmkc:
> It's working well on quest. The anchor spheres stay in the same place across recenter.

Glad to hear, worked well for me across different rooms too

dli7319 (approved):

whuang37:
Realigned into the new scriptmanager. all working for me

whuang37:
had to remove the virtualkeyboard for now due to the ui changes

ruofeidu:
> vlcsnap-2026-08-08-09h13m06s999
> It's not working on my Galaxy XR 😬. We'll need to check with Alex to see if it's implemented or not in Chrome.

I know this is still WIP for us.

Note: we are undergoing a major refactor before pushing new features / demos in today.

salmanmkc:
cool thanks! glad to know it's coming, would be surprised if anchors wouldn't be supported fully

salmanmkc:
> Realigned into the new scriptmanager. all working for me

great thanks!

dli7319:
Alex confirmed we have non-persistent spatial anchors.
https://immersive-web.github.io/webxr-samples/anchors.html works perfectly and stays in place across recenter on Galaxy XR.

I need to investigate the delta between your demo and the webxr sample.

PR #495, opened 2026-08-06
fix(virtualkeyboard): honour the keyboard transform

salmanmkc:
## Description

setting `position` on a `Keyboard` doesn't put it there. `init()` runs after your app has positioned it and moves the panel inside to a fixed spot, and that panel carries its own default placement underneath, so there are two things fighting whatever you set.

i ran into this building the netblocks integration sample and worked around it there by subclassing `Keyboard`, overriding `init()` and casting through to the panel it keeps private. that workaround is the clearest sign the api is wrong, so it goes away in this. `demos/math3d` has its own version of the same thing, a negative y picked to cancel out the offset that gets added afterwards.

moves the default onto the keyboard itself, set in the constructor, so assigning a position replaces it instead of being added to it, and pins the panel to the keyboard's origin.

a keyboard nobody places still appears exactly where it always did, `(0, 1.2, -1)` before and after. math3d lands in the same physical spot too, just written as where it goes rather than as an offset to cancel.

came up again while placing a keyboard in the spatial anchors demos on #494, where it landed a metre further away and high enough to sit through the panel above it. that demo compensates from the outside for now, and once this lands the compensation there is two constants to delete.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

- **Simulator Recording**: 
- **Device Recording**: 

## Checklist

- [x] **Tested in simulator & device**: verified in the desktop simulator. measured the keyboard's world position in math3d before and after, it lands in the same place, and a keyboard with no position set still defaults to `(0, 1.2, -1)` exactly as before.
- [x] **Large Assets ($\ge$ 1MB)**: none added.
- [x] **SDK Dynamic Dependencies**: no new dependencies.
- [x] **Security**: no keys or secrets.

### DIFF
--- a/demos/math3d/Math3D.js
+++ b/demos/math3d/Math3D.js
@@ -96,7 +96,7 @@ export class Math3D extends xb.Script {
 
     this.keyboard = new Keyboard();
     this.add(this.keyboard);
-    this.keyboard.position.set(0, -0.3, 0);
+    this.keyboard.position.set(0, 0.9, -1);
 
     const startFn = this.mathObjects[0].functionText;
 
--- a/src/addons/netblocks/samples/integration/main.ts
+++ b/src/addons/netblocks/samples/integration/main.ts
@@ -554,18 +554,10 @@ class IntegrationSample extends NetSample {
   }
 
   private _buildKeyboard(session: NonNullable<this['net']['session']>) {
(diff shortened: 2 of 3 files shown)


dli7319 (approved):
Good find. I've ran into a similar issue with panels in the past.

PR #498, opened 2026-08-06
New Templates!

whuang37:
# Updated Templates

## Description

- new templates and cleaned them up to better reflect current sdk
- some templates renamed just to better reflect a starting point 
- new ai prompt ui that handles the api keys a bit cleaner for browser usage via `options.ai.promptForApiKey = true;`
  - right now all the apps handle it slightly differently so its just a bit of cleanliness

- todo lockdown styling on this api key interface @ruofeidu 

- all templates tested on sim and device
## Type of Change

- [ ] Bug fix
- [x] New demo or sample
- [ ] Documentation update

image

- **Simulator Recording**: 
- **Device Recording**: 

## Checklist


src/ai/AIOptions.ts
```diff
@@ -26,6 +26,11 @@ export type AIModel = 'gemini' | 'openai';
 export class AIOptions {
   enabled = false;
   model: AIModel = 'gemini';
+  /**
+   * Show a browser dialog before AI starts so a prototype user can provide,
+   * replace, or remove a locally stored API key. Disabled by default.
+   */
+  promptForApiKey = false;
```
dli7319, line 35: @ruofeidu 
I know we didn't have this previously, but what do you think about enabling this by default?

src/ai/BrowserApiKeyPrompt.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  return new Promise((resolve) => {
+    let selectedConfiguredKey = configuredKey;
+    const displayName = modelName === 'openai' ? 'OpenAI' : 'Gemini';
+    const keyUrl =
+      modelName === 'openai'
+        ? 'https://platform.openai.com/api-keys'
+        : 'https://aistudio.google.com/app/apikey';
+    const dialog = document.createElement('dialog');
+    dialog.className = 'xb-ai-key-dialog';
```
dli7319: Can you make this a non-lit web component?
E.g. see the LoadingSpinner

dli7319 (approved):

src/ai/BrowserApiKeyPrompt.ts
dli7319: Or maybe just make it a lit web component and require lit for it.

src/ai/AIOptions.ts
dli7319: by default if enableAI() is called I mean.

dli7319:
Note that docs has some mechanism to cache an API key.
Please make sure that still works as well and doesn't cause an additional prompt.

whuang37:
> Note that docs has some mechanism to cache an API key. Please make sure that still works as well and doesn't cause an additional prompt.

it does it via local storage right? originally had that as the implementation but wasn't sure if we were chill with keeping api keys in local storage for all apps since technically you could just check the localstorage in there if there was a malicious actor

https://cheatsheetseries.owasp.org/cheatsheets/HTML5_Security_Cheat_Sheet.html

dli7319:
> > Note that docs has some mechanism to cache an API key. Please make sure that still works as well and doesn't cause an additional prompt.
> 
> it does it via local storage right? originally had that as the implementation but wasn't sure if we were chill with keeping api keys in local storage for all apps since technically you could just check the localstorage in there if there was a malicious actor
> 
> https://cheatsheetseries.owasp.org/cheatsheets/HTML5_Security_Cheat_Sheet.html

Looks like docs does it in its own local storage and then appends `key=` in the iframe url. Just make sure if `key=` is specified, the api key prompt doesn't appear.

src/ai/AIOptions.ts
whuang37: just added checks for keys.json and the url param so its safer to enable by default if we want to

ruofeidu (approved):

src/ai/AIOptions.ts
ruofeidu: I would prefer SDK provides a built-in way for prompt to reduce client-side code for AI prototypes.
I've asked for UX design.

templates/03_spatial_placement/main.js
```diff
@@ -0,0 +1,30 @@
...
+    this.add(new THREE.HemisphereLight(0xffffff, 0x3c4043, 3));
+    this.object = new THREE.Mesh(
+      new THREE.CylinderGeometry(0.09, 0.12, 0.22, 32),
+      new THREE.MeshStandardMaterial({color: 0x4285f4, roughness: 0.35})
+    );
+    this.object.name = 'Placed object';
+    this.object.position.set(0, -10, 0);
+    this.add(this.object);
+    void xb.world
+      .placeOnHorizontalSurface(this.object, {seconds: 15})
```
ruofeidu, line 15: why was 15 seconds chosen?

templates/05_camera_view/main.js
```diff
@@ -0,0 +1,53 @@
...
+      children: [this.preview],
+    });
+    card.position.set(0, 1.45, -1.2);
+    this.add(card);
+
+    this.camera = xb.core.deviceCamera;
+    this.onStateChange = (event) => this.updateState(event);
+    this.camera.addEventListener('statechange', this.onStateChange);
+    if (this.camera.state === 'streaming')
+      this.preview.src = this.camera.texture;
```
ruofeidu, line 37: if one line, do in the same line (not sure if lint will format to next)
otherwise, just quote it in {}

templates/10_environment_physics/EnvironmentProjectileDemo.js
```diff
@@ -0,0 +1,85 @@
...
+    if (this.projectiles.length > MAX_PROJECTILES)
+      this.removeProjectile(this.projectiles.shift());
+  }
+
+  physicsStep() {
+    const now = performance.now();
+    for (const projectile of [...this.projectiles]) {
+      projectile.mesh.position.copy(projectile.body.translation());
+      projectile.mesh.quaternion.copy(projectile.body.rotation());
+      if (now - projectile.bornAt > 12000 || projectile.mesh.position.y < -4)
```
ruofeidu, line 65: > 12000 /\*ms\*/

PR #502, opened 2026-08-07
feat(demos): pose detection on desktop with a real webcam

salmanmkc:
depends on #503 and #504.

the human pose detector demo does nothing in the simulator. no error, no skeleton, status just sits on "searching for user". `enableHumanDetection()` calls `enableCamera()` internally which defaults to the environment camera, and on desktop the simulator answers that one with a render of the virtual living room. so mediapipe is being handed a picture of a room that never has a person in it.

same fix `magic_window` already uses: `enableCamera('user')`, after the enable helper since it would overwrite it otherwise. the simulator camera has to stay registered though, it's the only thing supplying a camera pose on desktop, and with no pose `runDetectionInternal` bails and returns `[]` before the model runs. so turning off `simulator.deviceCamera.enabled` looks like the obvious fix and is actually a trap, you get webcam pixels and silently zero detections. gated on `isSessionSupported('immersive-ar')` so headsets keep the environment camera.

also a small camera preview in the corner so a camera that never starts stops looking identical to "no person in frame". mirrored with css, the frames going to mediapipe are deliberately left alone: landmark 11 is the subject's left shoulder, so mirroring the input swaps every left/right joint and gives you a plausible but wrong skeleton.

the skeleton placement this relies on, and the visibility filtering that stops it drawing guessed legs, are split into #503 and #504 so they can be reviewed on their own.

docs and the world sensing skill updated to match.


salmanmkc:
gonna split this into a stack

salmanmkc:
nevermind, I cannot create stacks on this repo as I only have read-only access to google/xrblocks . Since for stacked prs the base branch must live in the target repo. I will split it up into prs

whuang37:
Used this as a test to try migrating to new sdk version. migration lgtm + a bit extra ui styling :)

salmanmkc:
> Used this as a test to try migrating to new sdk version. migration lgtm + a bit extra ui styling :)

awesome, thank you William!

dli7319:
Nice!
For me, the 2D UI is showing in front of the skeleton lol.

Issue #373, opened 2026-06-15
perf: SDK-level hotspots surfaced from desktop sim traces


salmanmkc:
## context

while shipping #366 (move MediaPipe FaceLandmarker off the main thread) i started taking Chrome perf traces of demos that consume depth + interaction raycasting (face_mirror, portals, objects_3d). same SDK-level hotspots kept showing up. opened separate focused PRs for each, this issue collects them as one tracking thread.

all measurements are from chrome devtools profiler traces of `demos/face_mirror` (which uses depth + face detection) and `demos/portals` (depth + occlusion + interaction raycasting against 5 loaded scenes). times are main-thread self-time over a ~12 s sample window.

## pattern 1: simulator depth readback stalls main thread per frame

`SimulatorDepth.update()` fires a `readRenderTargetPixelsAsync` per frame. that helper polls a WebGL fence via `setTimeout`, which typically takes longer than a frame on desktop, so the polling chains stack up. accounted for ~5 s of every 12 s of main thread in the original face_mirror trace.

- fix: #370 (perf(simulator): don't stack depth-readback requests in flight)

## pattern 2: per-frame full-scene raycast is O(triangles)

the SDK's interaction raycaster (`Input.performRaycastOnScene`) walks every triangle of every mesh under `xb.core.scene` via `THREE.Raycaster.intersectObject` (recursive). for demos with non-trivial scene geometry (portals loads 5 immersive worlds) the per-triangle walk dominates the main thread. the per-landmark depth raycast in `processFaceLandmarkerResult` has the same shape.

three-mesh-bvh's accelerated raycaster fixes this but has to be opted into per geometry.

- fix: #372 (perf(utils): applyBVH helper for opting scenes into accelerated raycasts) - adds the SDK helper + opts portals in
- fix: #367 (perf(world/faces): BVH-accelerate per-landmark depth raycasts) - opts face_mirror in

## pattern 3: OcclusionPass setSize per frame triggers GPU texture realloc

`OcclusionPass` calls `setSize()` on its render targets every frame (occlusion map + 3 kawase blur targets). `setSize()` forces a GPU texture reallocation on every call even when dimensions are identical, and `renderer.getDrawingBufferSize` returns the same value frame after frame. ~154 ms in the portals trace.

- fix: #371 (perf(depth/occlusion): skip OcclusionPass setSize when dimensions unchanged)

## pattern 4: MediaPipe inference on main render thread

[...]

PR #505, opened 2026-08-07
skip the simulator depth readback when nothing has changed

salmanmkc:
follow-up to #373. pattern 1 there was the simulator depth readback stalling the main thread, and #370 fixed the part where the readback promises stacked up. it didn't change how often they run, and that turns out to be most of the desktop profile still.

`SimulatorDepth.update()` renders the depth scene and starts a readback every frame. the target is 160x160 single channel float, about 100 KB, so it isn't bandwidth, it's the stall: reading the target back waits on the GPU. measures at roughly 98 ms per readback on an intel UHD.

and in a stationary view every one of those comes back identical to the last. a probe over 120 consecutive frames showed the simulator camera doesn't move at all in automation mode, 0 frames changed.

so this skips the render + readback unless something that actually affects depth changed. camera transform, plus a hash over the world transform and visibility of every node the depth pass draws, so an object moving under a still camera still refreshes. `maxDepthAgeMs` bounds the worst case since a transform hash can't see skinning or vertex shaders.

numbers from `demos/portals` with a resting camera, five interleaved runs per side so both see the same thermal/background drift, 8s samples, mean ± sd, RTX 4060:

| metric | before | after |
|---|---|---|
| `getBufferSubData` | 244.5 ± 31.9 ms | 0.0 ± 0.0 |
| `getParameter` | 314.9 ± 33.8 ms | 0.0 ± 0.0 |
| `clientWaitSync` | 22.2 ± 3.7 ms | 0.0 ± 0.0 |
| main-thread idle | 67.2 ± 1.3 % | 77.7 ± 1.9 % |
| fps | 156.0 ± 4.5 | 154.6 ± 4.1 |

repeated the whole thing a second time at a different time/thermal state, same shape: `getBufferSubData` 375.9 ± 91.0 -> 1.9 ± 4.3, idle 59.8 ± 4.5 -> 71.8 ± 2.2, fps 151.4 ± 4.7 vs 150.6 ± 6.4.

**fps doesn't change.** the gain is main-thread headroom, ~580 ms of readback and fence-polling per 8s window handed back to app logic / AI / physics. i originally read a single-run pair as +14% fps and that was just noise, run to run spread is ±4.5.

on integrated graphics the readback is much more expensive (a single 10s run showed `getBufferSubData` 7705 ms -> 1721 ms and idle 16% -> 59%), which matters given XR devices are mobile-class GPUs. These are more indicative, since I didn't run these again over averages on the iGPU.

control: `samples/ui`, which has no depth, shows no `getBufferSubData` at all, so the cost really is the readback and not general scene load.

a resting camera is the best case. counting actual readbacks over 16s of moving around portals by hand: 73 would have run before, 33 ran after, so 55% fewer during real use rather than ~87% sitting still. the depth buffer itself comes out byte for byte identical either way, same checksum over all 25600 samples.

the hash itself is cheap, ~14 ms of a 9965 ms window.

10 unit tests cover camera move, object moving under a still camera, visibility toggle, object added, staleness refresh, and the existing in-flight guard.

### DIFF
--- a/src/simulator/scene/SimulatorDepth.test.ts
+++ b/src/simulator/scene/SimulatorDepth.test.ts
@@ -38,6 +38,28 @@ function makeMockRenderer() {
 describe('SimulatorDepth.update inflight guard', () => {
   let depthSim: SimulatorDepth;
   let renderer: ReturnType<typeof makeMockRenderer>;
+  let camera: THREE.PerspectiveCamera;
+  let simulatorScene: THREE.Scene;
+  let movingObject: THREE.Object3D;
+
+  /** Move the view so the depth buffer is genuinely out of date. */
+  const moveCamera = () => {
+    camera.position.x += 1;
+  };
+
+  /** Move something the depth pass draws, leaving the camera alone. */
+  const moveSceneObject = () => {
+    movingObject.position.z += 1;
+    movingObject.updateMatrixWorld(true);
+  };
+
+  /** Run update() and let the readback settle so the next call is unblocked. */
+  const settledUpdate = async () => {
+    depthSim.update();
+    renderer.settleReadback();
+    await Promise.resolve();
+    await Promise.resolve();
+  };
 
   beforeEach(() => {
     // jsdom doesn't ship XRRigidTransform; the readback path constructs
@@ -50,8 +72,12 @@ describe('SimulatorDepth.update inflight guard', () => {
         ) {}
       };
     renderer = makeMockRenderer();
-    const camera = new THREE.PerspectiveCamera();
-    depthSim = new SimulatorDepth({overrideMaterial: null} as never);
+    camera = new THREE.PerspectiveCamera();
+    simulatorScene = new THREE.Scene();
+    movingObject = new THREE.Object3D();
+    simulatorScene.add(movingObject);
+    simulatorScene.updateMatrixWorld(true);
+    depthSim = new SimulatorDepth(simulatorScene as never);
     depthSim.init(renderer.renderer as unknown as THREE.WebGLRenderer, camera, {
       updateCPUDepthData: vi.fn(),
     } as never);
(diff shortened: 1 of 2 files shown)


dli7319:
Thanks.
It's interesting that `getParameter` takes the longest time.
That seems to just be a sanity check and not doing anything meaningful.

Screenshot 2026-08-07 at 3 05 13 PM

dli7319 (approved):

dli7319:
I filed https://github.com/mrdoob/three.js/issues/34182.

If this gets addressed within threejs, readRenderTargetPixelsAsync is actually super fast:
Screenshot 2026-08-07 at 8 46 27 PM

dli7319:
>| metric | before | after |
|---|---|---|
| `getBufferSubData` | 244.5 ± 31.9 ms | 0.0 ± 0.0 |
| `getParameter` | 314.9 ± 33.8 ms | 0.0 ± 0.0 |
| `clientWaitSync` | 22.2 ± 3.7 ms | 0.0 ± 0.0 |
| main-thread idle | 67.2 ± 1.3 % | 77.7 ± 1.9 % |
| fps | 156.0 ± 4.5 | 154.6 ± 4.1 |

How can I get precise measurements like this?

salmanmkc:
It's a CDP script, nothing clever. Drives Chrome over the DevTools protocol, starts a trace, and counts the events straight out of it.

The ± comes from interleaving. Running A five times then B five times gives you thermal drift dressed up as a result, so it alternates A/B/A/B and reports mean and sd across the pairs. I had a single-run pair that looked like +14% fps and it was just noise, run to run spread is about ±4.5.

Worth checking what actually rendered, too. Chrome can quietly fall back to SwiftShader, which is software rasterisation on the CPU, and then you're timing the fallback instead of the GPU. The script reads the unmasked renderer string and warns if it sees it. I ran with `--force_high_performance_gpu` so it picks the discrete card rather than the integrated one, and `?xrAutomation=1` so the simulator starts without a click.

Happy to send it as a PR if that's useful, about 300 lines and no dependencies.

salmanmkc:
> I filed [mrdoob/three.js#34182](https://github.com/mrdoob/three.js/issues/34182).
> 
> If this gets addressed within threejs, readRenderTargetPixelsAsync is actually super fast: Screenshot 2026-08-07 at 8 46 27 PM

that's awesome, and nice pr you made there too! hope it gets merged soon

PR #507, opened 2026-08-07
skip the second eye's depth readback

salmanmkc:
`Depth.update` loops over the views in the frame, so on a stereo headset `updateGPUDepthData` runs once per eye. each call converts the GPU depth texture by rendering it and then calling the *synchronous* `readRenderTargetPixels`, which stalls the render thread until the GPU drains.

only the first view's CPU depth is ever consumed though. `getDepth`, `getDepthInMeters` and `getVertex` all read index 0 (`Depth.ts` lines 41, 130, 143, 165, 170, 188, 203) and the depth mesh update is already gated on `viewId == 0`. nothing in the SDK, demos, samples or templates reads `depthArray[1]` or `cpuDepthData[1]`. so in stereo half of those stalls produce data nothing looks at.

one condition:

```ts
const needCpuDepth = this.options.depthMesh.enabled && viewId === 0;
```

the per-view work that's actually needed stays: depth matrices are still computed for both eyes, and the native depth texture is still updated per view. neither of those stalls.

## how much is it worth

can't measure the real path without a headset, since it needs `depthUsage === 'gpu-optimized'` and a real `XRFrame`. so this is a proxy, measured without a laptop (RTX 4060, chrome 151) through the SDK's own renderer, on an R32F target the size of the XR depth buffer:

| target | render only | readback | total |
|---|---|---|---|
| 160x160 | 0.05 ms | 0.43 ms | 0.48 ms |
| 320x320 | 0.10 ms | 0.433 ms | 0.53 ms |

the readback is ~90% of it, and it's flat across a 4x pixel increase, which says you're paying for the pipeline stall rather than the pixels. removing one per frame is worth ~0.43 ms there, about 4% of a 90Hz budget.

to be clear those are on a discrete GPU, not device numbers, and not even the same code path. on a tile-based mobile GPU a readback forces tile memory out to main memory, so i'd expect the saving to be larger on device, but that's reasoning not measurement.

**would really appreciate someone with a Galaxy XR (or any Android XR device that reports `gpu-optimized` depth) running a before/after on this.** that's the one number i can't get.

## risk

`depthArray` and `cpuDepthData` are public fields, so an app reading index 1 directly would now see stale data instead of fresh. nothing in this repo does, and it's undocumented, but worth flagging.

three tests cover it: first view reads back, second view doesn't, and the per-view GPU data/matrices are still recorded for both. i checked the second-view test actually fails without the change rather than passing vacuously.

related: the same synchronous readback pattern is why #370 exists on the simulator side. `SimulatorDepth` and `ScreenshotSynthesizer` both use `readRenderTargetPixelsAsync`; `GPUDepthConverter` is the one remaining synchronous one. making it async would be the natural follow-up, but it trades the stall for a frame or two of latency plus the fence polling that #373 flagged, so halving the count first seemed like the cheaper win.

### DIFF
--- a/src/depth/Depth.test.ts
+++ b/src/depth/Depth.test.ts
@@ -143,4 +143,71 @@ describe('Depth', () => {
       vi.restoreAllMocks();
     });
   });
+  describe('updateGPUDepthData readback', () => {
+    /**
+     * Depth with just enough wired up to exercise the GPU depth path.
+     */
+    function createDepthWithConverter(convertGPUToCPU: () => unknown) {
+      const depth = createDepth();
+      depth.options = {
+        depthMesh: {enabled: true},
+        depthTexture: {enabled: false},
+      } as never;
+      (depth as unknown as {gpuDepthConverter: unknown}).gpuDepthConverter = {
+        convertGPUToCPU,
+      };
+      // updateDepthMatrices needs these populated per view.
+      vi.spyOn(
+        depth as unknown as {updateDepthMatrices: () => void},
+        'updateDepthMatrices'
+      ).mockImplementation(() => {});
+      return depth;
+    }
+
+    const fakeDepthData = {
+      width: 2,
+      height: 2,
+      data: new Float32Array([1, 2, 3, 4]).buffer,
+    } as unknown as XRWebGLDepthInformation;
+
+    it('reads back the first view, which is the one anything consumes', () => {
+      const convert = vi.fn(() => ({
+        width: 2,
+        height: 2,
+        data: new Float32Array([1, 2, 3, 4]).buffer,
+      }));
+      const depth = createDepthWithConverter(convert);
+
+      depth.updateGPUDepthData(fakeDepthData, 0);
+
+      expect(convert).toHaveBeenCalledTimes(1);
+      expect(depth.depthArray[0]).toBeInstanceOf(Float32Array);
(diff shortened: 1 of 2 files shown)


salmanmkc:
@dli7319 if you have a Galaxy XR (or any Android XR device that reports `gpu-optimized` depth) handy, would you mind running a before/after on this one?

the code path needs a real `XRFrame` plus `depthUsage === 'gpu-optimized'`, so it never executes in the desktop simulator and i had no way to measure the thing i'm actually claiming. the 0.43 ms in the description is with a discrete GPU, same API and format and buffer size but not the same path, so treat not indicative and I'd just treat that as direction-of-travel rather than a number.

what would settle it: a demo with `enableDepth()` on device, chrome devtools performance trace, look for `readRenderTargetPixels` self time before and after. if it turns out to be noise on Adreno then this isn't worth the small behaviour change and i'll close it.

dli7319:
Android XR only supports CPU depth so it won't be affected.

I'll check it on a Quest 3 later today.

dli7319:
Screenshot 2026-08-07 at 7 50 20 AM

Screenshot 2026-08-07 at 7 52 39 AM

Drops from ~5ms to ~3ms on Quest 3. Great find!

dli7319 (approved):

PR #508, opened 2026-08-08
feat(mcp): add a server exposing the xb-* skills and API surface

salmanmkc:
## Description

I've been meaning to do another eval round, but that comes out of my own API key, so I thought to improve it here before I run a whole sweep again.

`createXRScene()`, `useGesture()`, `xr.agent.hands.playGesture()`. None of those exist, they all came out of eval runs, and AGENTS.md already calls hallucinated APIs the top cause of broken generated apps.

To use it you point an MCP client at it, config is in `src/addons/mcp/README.md`. The client asks the server what tools it has on connect, so the model sees `list_skills`, `get_skill` and `search_api` in its context with a line each on when to reach for them. An agent picks `search_api` because it's the only thing that answers "is this symbol real", and `list_skills` because it's cheaper than reading 26 files to find out which one applies.

This doesn't replace the markdown though. In a clone the files are right there and AGENTS.md and GEMINI.md already point at them, so that case is already fine. it's when you've just installed the package that there's nothing, they're sat in node_modules and nothing tells an agent to go look. And checking whether a symbol is real has no answer either way, short of grepping the d.ts yourself.

So this is a stdio MCP server with three tools. `list_skills` and `get_skill` just read the markdown that already ships, so there's no second copy to drift. `search_api` looks a name up in `build/xrblocks.d.ts`, and when it doesn't match it says so outright, since "this isn't real, don't call it" is the answer worth having.

No runtime deps. The JSON-RPC framing is written out by hand rather than pulling in `@modelcontextprotocol/sdk`, because we've got one runtime dep today so I didn't want to increase that anymore. The catch is spec changes only turn up when some client refuses to talk to you, so the SDK is in as a devDependency and there's a test that drives the server with its actual client over a real stdio process: handshake, `tools/list`, a `tools/call` round trip, ping, unknown tool. Dev deps aren't installed by anyone installing xrblocks, so consumers still get zero. Bumping it now tells us if we've drifted instead of relying on someone reading a changelog. Checked it fails for the right reason by disabling ping, which the client rejects with -32601.

Most of the fiddly part was `search_api` being wrong..., since a lookup that says no to a real API is worse than no lookup. Indexing every declaration in the bundle is far too wide, it was happily confirming `sdk_Reticle` and the protected `Gemini.queryOnce`. Too narrow in the other direction as well, because option bags are declared inline, so `new ModelViewer({raycastToChildren})` came back fake. So i have made it such that it now filters through the trailing export statements and indexes members at any depth, skipping private and protected shapes. Checked 45 real APIs against 11 invented ones, nothing wrong either way.

`list_skills` shortens the descriptions, since it's the first call an agent makes and the whole thing lands in context before any work starts. Roughly 3.2k tokens down to 1.3k across 26 skills, and `get_skill` still gives you the full text.

There's a new workflow for this. `search_api` reads the generated d.ts, and whether that exists on CI depends on cache state: the test workflow only runs `npm ci` on a node_modules cache miss, and it's `npm ci` that triggers the `prepare` build. So this one builds explicitly, otherwise the check against the real file only happens on some runs, and a format change from the generator would go unnoticed with real names quietly no longer resolving.

**stdio only.** Works with anything that spawns a local process, so Gemini CLI, Copilot CLI, Claude Desktop, Cursor. Remote clients would want an HTTP transport, which isn't here.

## Type of Change

- [ ] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

Nothing to show, it's a dev-time server with no runtime or visual change.

## Checklist

- [x] **Tested in simulator & device**: Nothing in the SDK bundle changes, so not applicable to runtime. Tested by wiring it into an MCP client and calling all three tools, plus 43 tests over the handshake, notifications, error codes, malformed input, the symbol index, and a conformance pass driving the server with the official MCP client. Suite green at 669.
- [x] **Large Assets ($\ge$ 1MB)**: None.
- [x] **SDK Dynamic Dependencies**: No new runtime dependencies. `@modelcontextprotocol/sdk` is a devDependency used only by the conformance test, and rollup already ignores `src/addons/**/server/**` so none of this is bundled.
- [x] **Security**: No keys or secrets. `get_skill` matches names against the enumerated skill list rather than joining caller input onto a path, so it can't be walked out of `skills/`.

### DIFF
--- a/.github/workflows/mcp-check.yml
+++ b/.github/workflows/mcp-check.yml
@@ -43,9 +43,10 @@ jobs:
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: npm ci
 
-      # search_api reads the generated definitions, and the test workflow does
-      # not build. Without this, a format change in the generated .d.ts would
-      # silently stop real API names resolving and nothing would notice.
+      # search_api reads the generated definitions. The test workflow only
+      # builds them as a side effect of `npm ci`, which it skips on a
+      # node_modules cache hit, so the check against the real file would run
+      # only sometimes. Building here makes it deterministic.
       - name: Build SDK
         run: npm run build:sdk
 
--- a/.github/workflows/mcp-check.yml
+++ b/.github/workflows/mcp-check.yml
@@ -1,0 +1,53 @@
+name: MCP API Index Check
+
+on:
+  push:
+    branches-ignore:
+      - 'build'
+    paths:
+      - 'src/**'
+      - 'skills/**'
+      - 'package.json'
+      - 'package-lock.json'
+      - 'rollup.config.js'
+      - '.github/workflows/mcp-check.yml'
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  mcp-api-index:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
+
+      - name: Setup Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: '24'
+          cache: 'npm'
+
+      - name: Cache Node Modules
+        id: cache-node-modules
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6
+        with:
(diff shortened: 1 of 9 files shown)


dli7319:
Working in antigravity

Screenshot 2026-08-08 at 9 33 05 AM

salmanmkc:
> Working in antigravity
> 
> Screenshot 2026-08-08 at 9 33 05 AM

yay awesome!

PR #516, opened 2026-08-10
feat(layers): composition layer support behind enableLayers()

salmanmkc:
## Description

The device recording below was made on a Quest 3 against an earlier SDK build, with both the media and forced WebGL paths tested. The current demo has not been re-validated on a headset. Android XR hardware remains untested, including the dimensions and full behaviour of its WebGL path.

https://github.com/user-attachments/assets/2ad9d2b0-d0e9-469b-ac26-bf2800737f00

Anything drawn into the eye buffer is resampled at least twice before you see it: once into the buffer, which is already coarser than the panel across the area a given object covers, and again when the compositor warps it to head pose. A quad composition layer skips both and is sampled once, at its own resolution, straight by the compositor. That is a property of the render path rather than of any one headset.

The cost is easiest to see with numbers. A 1.2m quad two metres away covers about a third of the field of view, so on a Quest 3 it lands on roughly 620 eye buffer pixels across. Feed that a 2048 wide source and around 70% of the detail is gone before the compositor has even seen the frame. Fine text and single pixel lines are where it shows.

Adds `LayerManager`, `LayerCapability` and `VideoLayer` under `src/layers/`, plus a demo in `demos/layers/`. All behind `options.enableLayers()`, off by default.

After the spatial UI migration, the demo still called `SpatialPanel`, which no longer exists. That throws during script init, so the script never activates or reaches layer attachment. Replaces only the readout with `UICard` and `UIText`, plus the importmap entries that UI needs. The demo still creates its own manager; layer creation, video uploads and startup order are unchanged.

The same real Chromium check fails on `b0e15e19` with `SpatialPanel is not a constructor` and the script in `failedScripts`. After the readout fix, the script is active with no init exception. The card is constructed and mounted in the scene graph with the expected `UIText` values, both video meshes play, swaps update positions and text values, and the original `Object3D.layers` mask stays intact. This browser takes the unsupported-layer fallback path, not native composition layers.

Visible glyph rasterization could not be confirmed under headless SwiftShader. A separate control reproduces this pre-existing rasterizer limitation with an unmodified minimal public `UICard`: the same heading rectangle measures 35 bright pixels under SwiftShader versus 854 under real Metal. Headed SwiftShader also gives 35. The regression evidence here confirms initialization, scene-graph mounting and text values, not that the text was visibly rasterized.

Both backings are implemented, because neither one covers the devices we care about on its own. `XRMediaBinding` is the cheap path, where the compositor drives the video element and the app never draws a frame, but Quest is the only browser that ships it. Chrome has quad layers only through `XRWebGLBinding`, where the app uploads each frame itself, so that is the path Android XR needs.

`LayerManager` exists because three.js sets `layers: [projectionLayer]` once at session start and never touches that array again. Anything extra has to be composed back in together with the projection layer, and dropping it blanks the scene, so `setBaseLayer` is required before anything can be added.

`VideoLayer` reports whether it got a layer instead of throwing, so an app can ask for one everywhere and just draw into the scene where there isn't one. It falls back separately for no session, no base layer, no binding, and a platform that advertises a binding but refuses a particular element.

The quad sizing is worth calling out, because the two paths deliberately differ and it looks like a bug otherwise. The spec means `width` and `height` as full metres, and Chromium passes them to OpenXR unchanged, but the Quest compositor treats them as half extents and returns a quad at twice the size. It is a property of the compositor rather than of one binding, so the correction is decided by the platform, not the path. This is immersive-web/layers#324, still open. Worth knowing that three.js and A-Frame both hardcode a `/2` for Quest, which means copying either of them renders the quad at half size on Android XR.

A few other things the spec is explicit about and the official samples get wrong or omit: `layout` has to be named because `'default'` is a `TypeError` for quad layers, `isStatic` has to be false or `getSubImage` raises `InvalidStateError` once `needsRedraw` clears, and `getSubImage` throws until `updateRenderState` has taken effect a frame or two after the layer is added, so a failed upload drops that frame rather than tearing the layer down.

Two faults only showed up on device. The layer texture is allocated once at the size given at creation, and the test card is a canvas stream that reports no dimensions for its first frames, so guessing a size and then uploading a differently sized frame raised a GL error every frame and took the whole session down. And uploading through raw `gl.bindTexture` left three.js's state cache describing something untrue, so it carried on rendering with the compositor's texture bound. Uploads are also skipped for frames the video has not advanced past, since the source runs at 30fps against a 90fps render loop.

The demo puts the same generated card up twice, once as a layer and once as a plane in the scene, so the only variable is the path the frame takes. It is generated rather than a video file because the sample clips I tried are lower resolution than the eye buffer, so both sides looked identical and it showed nothing. The two sides swap on a timer, which matters: lenses are sharpest in the middle and you read one panel at a time, so a fixed left/right layout cannot separate "the layer is sharper" from "that side was in the sweet spot". Swapping does separate them, and on device the difference follows the layer across the swap. `?webglLayer=1` forces the WebGL path on a platform that has both, which is what lets a Quest stand in for Android XR.

## Type of Change

- [x] Bug fix
- [x] New demo or sample
- [ ] Documentation update

- **Simulator Recording**: the simulator has no compositor, so both sides fall back to scene textures and there is nothing to compare. The readout says so rather than leaving a blank space.
- **Device Recording**: filmed through the lens on a Quest 3 against an earlier SDK build (top of pr). The readout fix has not been retested on a headset.

## Checklist

- [x] **Tested in simulator & device**: Earlier SDK build tested on Quest 3 for both paths and in the desktop simulator for fallback. The current readout fix has real Chromium red/green initialization and scene-graph coverage, but no headset retest or visual glyph confirmation under SwiftShader. SDK build, lint and scoped Prettier pass; the full suite has 1,944 passed and one skipped. The 43 existing layer tests are preservation coverage, not new readout regressions. Android XR hardware remains untested.
- [x] **Large Assets ($\ge$ 1MB)**: None, the test card is generated at runtime onto a canvas.
- [x] **SDK Dynamic Dependencies**: No SDK dependencies added. The demo importmap now includes the existing dependencies required by its replacement UI.
- [x] **Security**: No keys or secrets.


dli7319:
Both sides are looking the same to me on Galaxy XR. Hmm.
Maybe composition layers is behind a flag or something in Chrome.
I need to ask around.

salmanmkc:
> Both sides are looking the same to me on Galaxy XR. Hmm. Maybe composition layers is behind a flag or something in Chrome. I need to ask around.

ah, thanks for testing. it might be falling back to a normal scene texture on both sides, which would explain why they look the same. does the in-headset readout show `webgl | attached: true` with uploads increasing, or `none | attached: false`?

dli7319:
I plan to submit this but I probably won't have time to figure out why it doesn't work on Galaxy XR.
I'll submit it once I test it on Quest 3.

salmanmkc:
> I plan to submit this but I probably won't have time to figure out why it doesn't work on Galaxy XR. I'll submit it once I test it on Quest 3.

cool thanks! once Aura releases later this year I'll probably see if I can figure it out

dli7319:
I loaded it up on quest. The layers part is working but the page crashes after a minute or two.

Screenshot 2026-09-19 at 9 43 02 PM

dli7319:
Longer trace:

Screenshot 2026-09-19 at 9 49 12 PM
[Trace-20260919T214539.json.gz](https://github.com/user-attachments/files/32429417/Trace-20260919T214539.json.gz)

demos/layers/LayersScene.js
```diff
@@ -0,0 +1,305 @@
...
+    // WebGL one has nowhere to run without being asked for. Same flag lets a
+    // Quest stand in for Android XR, which only has the WebGL path.
+    this.requestedWebGL = xb.getUrlParamBool('webglLayer', false);
+    this.manager.setPreferWebGL(this.requestedWebGL);
+
+    // One generated card feeding both copies, so the only variable is which
+    // path the content takes to the display.
+    this.card = new TestCard();
+    this.layerVideo = this.card.start();
+    this.sceneVideo = this.card.start();
```
dli7319, line 51: My antigravity is saying calling start() twice causes another loop which never gets cleaned up.

src/layers/VideoLayer.ts
```diff
@@ -0,0 +1,357 @@
...
+  video: HTMLVideoElement,
+  session: XRSession,
+  space: XRReferenceSpace,
+  transform: XRRigidTransform,
+  width: number,
+  height: number
+): XRQuadLayer | null {
+  const MediaBinding = (
+    globalThis as {
+      XRMediaBinding?: new (session: XRSession) => {
```
dli7319: Can you get this PR'ed into the webxr types?

dli7319:
Also my agy says we shouldn't keep repeatedly calling `createQuadLayer`.

src/layers/VideoLayer.ts
salmanmkc: checked this and `XRMediaBinding` / `XRMediaQuadLayerInit` are already in `@types/webxr@0.5.24`, which our lockfile uses. de11554a removes the handwritten shape and uses those declarations directly, keeping the runtime check for browsers without the binding. no upstream addition needed.

demos/layers/LayersScene.js
salmanmkc: yep, both calls were starting their own drawing loop but overwriting the same timer handle. fixed in 165ce631: the two video elements now share one drawing loop, and cleanup cancels it, stops both streams, removes the videos and disposes the meshes. added regressions for two consumers, repeated cleanup and session re-entry.

salmanmkc:
thanks for the Quest report and trace. fixed the duplicate drawing loop and changed the swaps to update the existing quad transform instead of calling `createQuadLayer` again every three seconds. also bounded attachment attempts and cleaned up streams and layers on teardown/failure (165ce631).

084ee36f also aligns the demo with three.js 0.186.0. the old pin caused a reproducible cleanup error in the browser.

i have been testing the updated preview on my Quest 3 for a couple of minutes without a crash so far. encouraging, but not claiming a confirmed long-run fix or that both the media and forced-WebGL paths have been retested yet. -

demos/layers/index.html
```diff
@@ -0,0 +1,46 @@
...
+        "imports": {
+          "three": "https://cdn.jsdelivr.net/npm/three@0.186.0/build/three.module.js",
+          "three/addons/": "https://cdn.jsdelivr.net/npm/three@0.186.0/examples/jsm/",
+          "three/": "https://cdn.jsdelivr.net/npm/three@0.186.0/",
+          "@pmndrs/uikit": "https://cdn.jsdelivr.net/npm/@pmndrs/uikit@1.0.64/dist/index.min.js",
+          "@pmndrs/uikit-pub-sub": "https://cdn.jsdelivr.net/npm/@pmndrs/uikit-pub-sub@1.0.64/dist/index.min.js",
+          "@pmndrs/msdfonts": "https://cdn.jsdelivr.net/npm/@pmndrs/msdfonts@1.0.64/dist/index.min.js",
+          "@preact/signals-core": "https://cdn.jsdelivr.net/npm/@preact/signals-core@1.14.0/dist/signals-core.mjs",
+          "yoga-layout/load": "https://cdn.jsdelivr.net/npm/yoga-layout@3.2.1/dist/src/load.js",
+          "troika-three-text": "https://cdn.jsdelivr.net/gh/protectwise/troika@028b81cf308f0f22e5aa8e78196be56ec1997af5/packages/troika-three-text/src/index.js",
```
dli7319: Can you remove the troika imports including bidi-js and webgl-sdf-generator?

dli7319:
Just retested and it works fine now after 2 minutes. Thank you

Still the same interesting memory graph on Quest tho.

demos/layers/index.html
dli7319: I'll just push it myself since it's trivial

salmanmkc: cool thanks!

salmanmkc:
> Just retested and it works fine now after 2 minutes. Thank you
> 
> Still the same interesting memory graph on Quest tho.

I'll try look into the graph a bit more, thanks!

PR #523, opened 2026-08-11
Fix anchors on Galaxy XR

dli7319:
## Description

- Remove frame caching in AnchorManager and Core, obtaining frames directly via renderer.xr.getFrame()
- Introduce XRReferenceSpaceCache to cache reference spaces on session start and perform pose conversions
- Update AnchorManager.create() to default anchorSpace to 'bounded-floor' and transform poses when source and target spaces differ
- Request local-floor, bounded-floor, and unbounded reference spaces by default in WebXR session options
- Update tests and demo scripts to match updated reference space and frame querying logic

## Type of Change

- [x] Bug fix
- [x] New feature / enhancement

## Media / Screen Recordings & Screenshots (If Applicable)

https://github.com/user-attachments/assets/ab41af5b-fbc7-4b54-bccb-96b03ddd2259

## Checklist

- [x] **Tested in simulator & device**: Verified functionality in desktop simulator and/or physical hardware (where applicable).
- [x] **Large Assets ($\ge$ 1MB)**: Submitted separately to [xrblocks/proprietary-assets](https://github.com/xrblocks/proprietary-assets) via jsdelivr CDN.
- [x] **SDK Dynamic Dependencies**: All new SDK dependencies are dynamically loaded at runtime.
- [x] **Security**: Confirmed no hardcoded API keys or secrets are committed.

PR #524, opened 2026-08-11
fix(anchors): restore anchor creation and persistence 

salmanmkc:
## Description

Follow-up to #523, adjusting a few paths for the move away from a cached `XRFrame`.

| Fix | Why |
| --- | --- |
| `create()` queues when there is no live frame | `getFrame()` is null outside the animation callback, so nothing was placed when called from a click handler |
| `restoreAll()` falls back to the renderer's session | Keeps the existing no-argument callers working |
| Anchor space falls back when `bounded-floor` isn't granted, warns once | Hardware without a configured play space could not anchor |
| A named `poseSpace` that can't be resolved fails rather than guessing | Substituting a space would misplace the anchor |
| Reference spaces resolving after their session ended are ignored | A late promise could repopulate a cleared cache |
| `getPose()` guards again and records `lastError` | Can throw across sessions, and it's called every frame |
| `Core.currentFrame` no longer throws before the renderer exists | Reading it early gave a TypeError rather than undefined |
| `Core` keeps `referenceSpaceType` in the optional features | An app setting `webxrOptionalFeatures` itself dropped `local-floor` |

`bounded-floor` stays the default, it just falls back now rather than declining to anchor. Placement is unaffected either way, as the pose is converted into whichever space is used.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

Checked on a Quest 3 with this branch and `main` served side by side, same demo, only the SDK bundle differing.

## Checklist

- [x] **Tested in simulator & device**: Quest 3, compared against `main`. 619 unit tests pass. The one full-suite failure (`Core.test.ts` simulator lifecycle) also reproduces on `main` and is unrelated.
- [x] **Large Assets ($\ge$ 1MB)**: None.
- [x] **SDK Dynamic Dependencies**: None added.
- [x] **Security**: No keys or secrets.

### DIFF
--- a/src/core/Core.ts
+++ b/src/core/Core.ts
@@ -564,8 +564,8 @@ export class Core {
       this.context.setDeviceCamera(this.deviceCamera);
     }
 
-    const webXRRequiredFeatures: string[] = [...options.webxrRequiredFeatures];
-    const webXROptionalFeatures: string[] = [...options.webxrOptionalFeatures];
+    const webXRRequiredFeatures: string[] = options.webxrRequiredFeatures;
+    const webXROptionalFeatures: string[] = options.webxrOptionalFeatures;
     // The space the scene is drawn in has to be granted, and an app that sets
     // webxrOptionalFeatures itself would otherwise drop it and fail to start.
     if (!webXROptionalFeatures.includes(options.referenceSpaceType)) {
--- a/src/core/Core.ts
+++ b/src/core/Core.ts
@@ -564,8 +564,13 @@ export class Core {
       this.context.setDeviceCamera(this.deviceCamera);
     }
 
-    const webXRRequiredFeatures: string[] = options.webxrRequiredFeatures;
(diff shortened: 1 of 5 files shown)


src/core/Core.ts
```diff
@@ -564,8 +564,13 @@ export class Core {
       this.context.setDeviceCamera(this.deviceCamera);
     }
 
-    const webXRRequiredFeatures: string[] = options.webxrRequiredFeatures;
-    const webXROptionalFeatures: string[] = options.webxrOptionalFeatures;
+    const webXRRequiredFeatures: string[] = [...options.webxrRequiredFeatures];
```
dli7319: Nit: Can we keep the `options` instead of cloning it?

dli7319 (approved):
Tested on galaxy xr.

src/core/Core.ts
salmanmkc: yup! thanks for updating

PR #543, opened 2026-08-20
Pose skeleton draws over the head-locked HUD card

salmanmkc:
# Pose skeleton draws over the head-locked HUD card

## Description

`demos/human_pose_detector/` — noticed on #502. the HUD card is head-locked 0.8 m out while the skeleton lands 2 m out (`METRIC_SKELETON_DISTANCE_METRES`), so the card sits between you and the body it is describing and washes the cyan lines out.

nothing is actually occluding it. every visible mesh of the card is transparent with `depthWrite: false`, so no depth is ever written at 0.8 m. the joint markers were opaque though, and three draws the entire opaque group before any transparent geometry regardless of `renderOrder`, so the skeleton could never sort above the card's panel layers.

moves the markers into the transparent pass and gives markers and connectors a `renderOrder` of 1000, in line with the 999-1002 already used in `unistroke_recognizer` and `navmesh_simulator`. `depthTest` and `depthWrite` are left alone, so the skeleton still self-occludes and opaque scene geometry still hides it.

measured on the most saturated cyan pixel inside the card footprint: `rgb(102,155,166)` before, `rgb(13,235,250)` after, which is the connector colour `0x00f0ff` essentially unblended.

worth flagging for review: this only moves about 1.6% of the card's pixels, because the connectors are `0.003` radius at 2 m so it is thin lines rather than a dramatic reveal. if that still reads as too subtle, the alternative is a layout change instead, nudging the `FollowHead` offset so the card clears the body outright.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

- **Simulator Recording**: 
- **Device Recording**: not tested on hardware, desktop simulator only

## Checklist


ruofeidu:
Thank you Salman! Can you share a screenshot in the simulator? This one simulator alone might be fine. I can test on device tomorrow.

dli7319 (approved):

PR #571, opened 2026-09-10
feat(addon): Roomcraft scene authoring with Gemini

salmanmkc:
## Description

Roomcraft generation example

describe a place and get actual 3D objects, then keep refining the same scene with prompts or direct manipulation. Roomcraft is an opt-in add-on in `src/addons/roomcraft/`, with the interactive demo in `demos/roomcraft/`.

composes catalog assets and bounded procedural designs, including articulated parts that swing or spin. optional virtual-world mode gives the scene its own ground, sky, lighting, ponds, paths and seeded plantings. this is procedural composition, not a text-to-mesh service or generated JavaScript.

desktop and spatial studios share a draft, selection, undo/redo, surface placement and JSON export. voice uses the existing Gemini key, with cancellation and draft protection, no second speech provider. invalid plans can get one correction attempt before anything is applied; failed requests keep the current scene intact.

actually now i've merged in main since those have all merged in separately now as fixes but originally I had them here:
also includes shared SDK fixes that surfaced while using the demo: typing no longer moves the simulator, XR entry shows failures and stays retryable, scripts get the current headset camera pose, and stale hand touches stop blocking controller rays. the Quest virtual-mode change that omits optional `unbounded` is demo-only.

## Type of Change

- [ ] Bug fix
- [x] New feature / enhancement
- [x] New demo or sample
- [x] Documentation update

## Media / Screen Recordings & Screenshots (If Applicable)

Meta Quest and Simulator recording (4:05 is where simulator starts, Quest from the beginning): https://youtu.be/0CXubRowyPo

- [x] **Tested in simulator & device**: Desktop simulator and Quest runs covered scene generation and refinement. Quest runs also covered XR entry, Gemini voice and invalid-plan correction. Device confirmation of the latest stale-touch fix is still pending.
- [ ] **Large Assets ($\ge$ 1MB)**: N/A, no large assets added to the repository. The optional glTF exhibit uses an existing CDN-hosted model.
- [ ] **SDK Dynamic Dependencies**: N/A, no new npm dependencies. Reuses the existing SDK and add-ons.
- [x] **Security**: No hardcoded API keys or secrets committed. Prototype keys stay in page memory; production AI calls still need a server-side proxy.

PR #573, opened 2026-09-10
show XR entry failures and keep retries safe

salmanmkc:
## Description

hit Enter XR, the request gets rejected, and nothing happens. the button keeps saying ENTER XR, no message anywhere, and clicking again looks exactly like the first click. the only way to find out what went wrong is the devtools console.

`WebXRSessionManager` now dispatches a `SESSION_ERROR` event carrying the original error, holds the pending-request guard until renderer init actually finishes instead of dropping it as soon as `requestSession` resolves, and ends a session whose `renderer.xr.setSession` threw so the device isn't left holding one nothing owns. if that close also fails, both errors come back as an `AggregateError`. `XRButton` shows `ENTERING XR...` while a request is in flight, surfaces the failure in a `role="alert"`, and puts the button back in a usable state after a permission denial, a rejected request, or a failed setup.

extracted from #571 so it can land on its own.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

## Checklist

- [ ] **Tested in simulator & device**: no manual simulator or device run was performed. covered by the new `XRButton.test.ts` (jsdom). `npm test` (668 pass), `npm run lint`, prettier, and `npm run build:sdk` are all clean.
- [ ] **Large Assets ($\ge$ 1MB)**: N/A, no assets added.
- [ ] **SDK Dynamic Dependencies**: N/A, no new dependencies.

### DIFF
--- a/src/core/components/WebXRSessionManager.ts
+++ b/src/core/components/WebXRSessionManager.ts
@@ -6,6 +6,7 @@ export enum WebXRSessionEventType {
   READY = 'ready',
   SESSION_START = 'sessionstart',
   SESSION_END = 'sessionend',
+  SESSION_ERROR = 'sessionerror',
 }
 
 export type WebXRSessionManagerEventMap = THREE.Object3DEventMap & {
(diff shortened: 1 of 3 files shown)


salmanmkc:
This would happen for example if the device doesn't support a certain feature, such as plane detection but it wouldn't be known why it would fail

dli7319 (approved):

PR #574, opened 2026-09-10
refresh the headset camera before scripts run

salmanmkc:
# Pull Request Template

## Description

on XR entry the public camera still holds the pose from before the session started, so input and scripts run against a stale headset pose for that frame while rendering updates it afterwards. anything that reads the camera to place view-space UI gets the wrong pose, and any frame where the render-time update is the only refresh has the same ordering problem.

`Core.update` already pulled the current pose via `renderer.xr.updateCamera`, but only when `effects` was set, since XREffects renders each eye manually with three's auto camera update off. without postprocessing the public camera only gets refreshed during render, which is after input sampling and script updates have already run.

drops the effects condition so the pose refresh happens whenever we're presenting. the effects path and non-XR behavior are unchanged. new Core test runs both with and without postprocessing and checks a script's `update` observes the headset position, it fails on the no-postprocessing case without the fix.

extracted from #571 so it can land on its own.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

## Checklist

- [ ] **Tested in simulator & device**: N/A, no manual simulator or device run for this one. the new Core regression passes both with and without postprocessing, and fails on the no-postprocessing case when the fix is reverted. separately, a default `npm test` run here also hits one unrelated failure in `Core.test.ts > shares one in-flight simulator start and ignores later starts once running`, which reproduces on pristine main at the same base with this change reverted, so it is a baseline flake rather than anything from this fix.
- [ ] **Large Assets ($\ge$ 1MB)**: N/A, no assets in this change.
- [ ] **SDK Dynamic Dependencies**: N/A, no new SDK dependencies.

### DIFF
--- a/src/core/Core.test.ts
+++ b/src/core/Core.test.ts
@@ -170,6 +170,36 @@ describe('Core frame and simulator lifecycle', () => {
     expect(core.renderer.render).toHaveBeenCalledWith(core.scene, core.camera);
   });
 
+  it.each([false, true])(
+    'updates the headset camera before scripts (postprocessing: %s)',
+    async (postprocessing) => {
+      if (postprocessing) {
+        core.effects = {
+          render: vi.fn(),
+          dispose: vi.fn(),
+        } as unknown as Core['effects'];
+      }
+      const observedPosition = new THREE.Vector3();
+      const script = new Script();
+      vi.spyOn(script, 'update').mockImplementation(() => {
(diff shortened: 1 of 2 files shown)


dli7319 (approved):

salmanmkc:
thanks @dli7319!

PR #576, opened 2026-09-10
don't drive the simulator while a text field has focus

salmanmkc:
## Description

typing into a text field in the desktop simulator also drives the simulator. `w`/`a`/`s`/`d` walk the camera while you type, `1` and `2` flip the stereo render mode, and backtick toggles the UI. and a movement key you were already holding when the field took focus stays held, so the camera keeps moving while you type into the field.

`SimulatorControls.onKeyDown` now bails when the composed path or `document.activeElement` is an input, textarea, select or contenteditable, so text controls inside open shadow roots are covered too. IME composition keys are skipped for the same reason. a `focusin` listener clears held navigation keys when you click into a field so movement stops instead of sticking, and it comes back off on disconnect.

keyup is deliberately left alone so a key released over a text field still clears. leaving a field doesn't resume movement until you press a navigation key again. normal text editing and its key events aren't blocked.

some apps guard this by hand, e.g. the aisimulator panel swallows every keydown on its text input with `stopPropagation()` purely so the simulator doesn't see it. plenty don't: netblocks' chat box and room-code field are plain inputs with no guard, so typing a room code today also steers the camera. the hand-rolled guards still work and aren't broken by this, but neither kind of app should need one now that the controls check focus themselves.

pulled out of #571 so it can land on its own ahead of the larger feature.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [x] Documentation update

## Checklist

- [ ] **Tested in simulator & device**: Verified functionality in desktop simulator and/or physical hardware (where applicable).
- [ ] **SDK Dynamic Dependencies**: All new SDK dependencies are dynamically loaded at runtime.

N/A for large assets (no assets added) and for SDK dynamic dependencies (no new dependencies), so those two are left unchecked rather than asserted.

no manual simulator or device run, so that box is unchecked too. the behavior is covered by unit tests in `SimulatorControls.test.ts` instead: each field type, shadow-root inputs, editable-region descendants, keys dispatched on the document while a field holds focus, the held-key clear on focus, keyup still clearing, IME composition, and listener removal on disconnect. that file is 12/12 green, and `eslint src`, prettier on the changed files, and `rollup -c --failAfterWarnings` are all clean.

one unrelated test fails in the default full run, `src/core/Core.test.ts > Core frame and simulator lifecycle > shares one in-flight simulator start and ignores later starts once running`. it reproduces the same way on unchanged main and passes when that file runs on its own. capping the pool with `--maxWorkers=2` also gets the whole suite green, so it's a pre-existing concurrency flake rather than something this branch introduces. left alone.

### DIFF
--- a/docs/docs/manual/Simulator.mdx
+++ b/docs/docs/manual/Simulator.mdx
@@ -80,6 +80,8 @@ options.simulator.modeToggle.enabled = true;
 When mode toggling is enabled, the configured toggle key cycles through
 `options.simulator.modeToggle.toggleOrder`.
 
+Keyboard navigation and simulator shortcuts are suspended while an input, textarea, select, or editable region has focus, including text controls inside open shadow roots. Focusing a text control clears held movement keys; leaving it does not resume movement until another navigation key is pressed. Text editing and its normal key events are not blocked, and IME composition does not trigger simulator shortcuts.
+
 ## Automation preset
 
 Use the shared preset for automated or external browser runs:
--- a/src/simulator/SimulatorControls.test.ts
+++ b/src/simulator/SimulatorControls.test.ts
@@ -1,4 +1,4 @@
-import {afterEach, describe, expect, it, vi} from 'vitest';
+import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest';
 
 import {SimulatorControllerState} from './SimulatorControllerState';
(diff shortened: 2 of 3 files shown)


dli7319:
Thank you!

dli7319 (approved):

PR #577, opened 2026-09-12
perf(context): skip redundant transform refreshes in semantic snapshots

salmanmkc:
## Description

`buildSemanticTree` already calls `scene.updateMatrixWorld(true)` once before it traverses the scene. Despite that, `createSemanticNode` called `object.updateMatrixWorld(true)` again for every node it visited, which forced a full recursive refresh of that node's subtree. In a deep hierarchy the same descendants get refreshed once per ancestor, so the cost grows with depth while the matrices are already current.

This removes that per-node call. The scene-wide refresh at the start of the snapshot stays exactly where it is, and `object.getWorldPosition(tempPosition)` stays too, since it performs the ancestor walk the node actually needs. Bounds are unaffected because `getObjectBounds` / `getUIObjectBounds` refresh their own subtrees, and `Box3.expandByObject` updates each object it visits.

Keeping `getWorldPosition` matters for correctness, not just style. `UIElement` overrides it so a UI node reports the world position of its registered uikit presentation object rather than its logical transform. Reading `matrixWorld` directly instead, as a first pass did, silently reports the wrong position for those elements once the logical and presentation transforms diverge.

There is no caching across frames or snapshots, no change to camera or XR pose timing, and no public API change.

On performance, this is a redundant-work cleanup rather than a headline win. An isolated synthetic benchmark over a deep semantic snapshot showed a small, repeatable reduction in per-snapshot build time. Semantic context polling defaults to 3000 ms, so this is not expected to move frame rate in a typical app.

Covered by a new colocated `SemanticTreeBuilder.test.ts`:

- positions and bounds for objects beneath transformed, rotated, and scaled parents
- fresh values on a later snapshot after objects move and rescale
- no forced per-node subtree refresh during a snapshot
- a custom `getWorldPosition` override remains authoritative
- a real `UICard` / `UIPanel` registered through the presentation helper under a separate render root, where the logical and presentation transforms differ

Three of those cases fail against the behavior they guard, so they are live regressions rather than restatements.

Verified in Chromium as well as in unit tests: the same page was loaded against a build with and a build without this change, and the resulting semantic trees, visible-object contexts, and bounds were byte-identical across three snapshots taken before and after moving, rotating, and scaling both plain meshes and spatial UI.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

- **Simulator Recording**: not applicable, there is no visible behavior change. Verified in the desktop simulator by comparing semantic snapshots before and after the change.
- **Device Recording**: not applicable.

## Checklist

- [x] **Tested in simulator & device**: Verified in the desktop simulator. Not tested on physical hardware.

### DIFF
--- a/src/context/scene/semantic-tree/SemanticTreeBuilder.test.ts
+++ b/src/context/scene/semantic-tree/SemanticTreeBuilder.test.ts
@@ -1,0 +1,167 @@
+import * as THREE from 'three';
+import {afterEach, describe, expect, it, vi} from 'vitest';
+
+import {registerUIPresentationObject} from '../../../ui/UIElement';
+import {UICard} from '../../../ui/components/UICard';
+import {UIPanel} from '../../../ui/components/UIPanel';
+import {SemanticIdRegistry} from '../../shared/SemanticIdRegistry';
+import {buildSemanticTree} from './SemanticTreeBuilder';
+
+function findNode(
+  tree: ReturnType<typeof buildSemanticTree>,
+  name: string
+): (typeof tree.tree.nodes)[string] | undefined {
+  return Object.values(tree.tree.nodes).find((node) => node.name === name);
+}
+
+describe('buildSemanticTree', () => {
+  afterEach(() => vi.restoreAllMocks());
+
+  it('reports world positions and bounds under a transformed parent', () => {
+    const scene = new THREE.Scene();
+    const parent = new THREE.Group();
+    parent.name = 'Parent';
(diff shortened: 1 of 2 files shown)


salmanmkc:
this is actually just removing one line with a bunch of tests

dli7319:
Can you fix the merge conflict?

salmanmkc:
> Can you fix the merge conflict?

Yup done!

dli7319:
Thanks.

PR #578, opened 2026-09-12
reuse obstacle bounds after a rejected placement candidate

salmanmkc:
## Description

`placeOnHorizontalSurface` walks the same obstacle geometry again for every candidate it rejects in a crowded scene.

Keeps the first candidate on one scratch box, then caches obstacle bounds if it fails. The cache resets after each frame yield, and bounds containing the moving object are never cached. Collision testing, scoring, transform restoration and timeout behaviour are unchanged, and there is no public API or configuration change.

In an isolated controlled Chrome 152 / Apple M4 benchmark with 56 candidates and 151 obstacle meshes, synchronous search-pass time went from 3.332 +/- 0.232 ms to 0.338 +/- 0.034 ms across five interleaved pairs. No clear change when the first choice was free. This is less search work, not an FPS or full placement-latency claim.

Checked in the desktop simulator with a cluttered surface: the same spot is chosen as before the change, with obstacle bounds calculations falling from 839 to 105 for that placement.

Colocated coverage includes first-choice scratch reuse with nothing cached, bounded recomputation, moved obstacles after a yield, ancestor bounds never being cached, and restoring the target after a failed pass.

## Type of Change

- [ ] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

- **Simulator Recording**:  Not attached; no visual change intended.
- **Device Recording**:  Not attached; no physical-device run.

## Checklist

- [ ] **Tested in simulator & device**: Automated geometry cases plus a scripted desktop simulator run comparing placement before and after; no physical-device run.
- [x] **Large Assets ($\ge$ 1MB)**: Submitted separately to [xrblocks/proprietary-assets](https://github.com/xrblocks/proprietary-assets) via jsdelivr CDN. N/A, no assets added.
- [x] **SDK Dynamic Dependencies**: All new SDK dependencies are dynamically loaded at runtime. N/A, no dependencies added.

### DIFF
--- a/src/world/HorizontalPlacement.test.ts
+++ b/src/world/HorizontalPlacement.test.ts
@@ -1,0 +1,164 @@
+import * as THREE from 'three';
+import {afterEach, describe, expect, it, vi} from 'vitest';
+
+import {WaitFrame} from '../core/components/WaitFrame';
+
+import {placeOnHorizontalSurface} from './HorizontalPlacement';
+import {DetectedPlane} from './planes/DetectedPlane';
+import {PlaneDetector} from './planes/PlaneDetector';
+
+function createPlacement() {
+  const scene = new THREE.Scene();
+  const camera = new THREE.PerspectiveCamera();
+  camera.position.y = 1.6;
+  const planes = new PlaneDetector();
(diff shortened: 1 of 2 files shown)


dli7319:
Thanks!

Issue #569, opened 2026-09-08
Change raycast line default to something more reasonable.


dli7319:

salmanmkc:
a few ideas in case useful here @dli7319 

| idea | what's there now | thoughts |
|---|---|---|
| softer line | fixed 5-metre line, default material | lower opacity or a fade might be enough to make it less intrusive. shortening it alone could leave a gap between the line and a distant reticle. |
| stop at the hit | the reticle uses the hit position, but the line doesn't | could use the same hit, no extra raycast. would need a fallback for `raycastMode: 'select'`, since there isn't an idle hit in that mode. |
| fade or hide when pointing at nothing | the reticle hides on a miss; the line stays the same length | a faint guide might help with aiming, while hiding it would be cleaner. probably depends on the intended feel. |
| hover / press feedback | the reticle already has a hover ring and pressed state | the line could echo that, though it might be redundant if the reticle is already clear enough. |
| hide during direct touch | touch already suppresses ray targeting and hides the reticle, but the line doesn't follow that state | seems worth considering so the line isn't suggesting ray interaction while touch is active. |
| different width or taper | currently a plain `THREE.Line` | could look nicer, but consistent width/taper across devices may need different geometry or shaders. more involved than an opacity change. |

dli7319:
We're doing some benchmarking and it seems every generated app which has a THREE.Line ends up breaking the hand raycast.
I think it's this parameter we need to change: https://threejs.org/docs/#Raycaster.params

salmanmkc:
ahh, I was thinking of the hand ray visual, my bad 

yeah, looks like it's using three's default threshold of `1` world unit. That may beretty big hit area for a thin line, so that could explain it grabbing hits meant for something behind it.

the netblocks sample has a `0.0001` workaround too, although that's using the old raycaster API.

salmanmkc:
follow-up on the netblocks one: #585 drops that `0.0001` workaround, since `input.raycaster` isn't around anymore. the edge lines get opted out of hit testing instead, `edges.xb = {pointerEvents: 'none'}`, which is what `HitRegistry` reads.

that only works because the sample knows to mark the object though. a generated app that drops in a `THREE.Line` won't, and it'll still swallow hits at the default threshold of 1. so the per-object opt-out isn't really a substitute for changing the default here.

salmanmkc:
had a dig into this one and wrote a couple of throwaway tests to check I wasn't talking nonsense.

[...]

PR #585, opened 2026-09-14
fix(netblocks): move addon samples to demos so they build and run again

salmanmkc:
## Description

the netblocks and lipsync sample pages 404 on their entry script, so nothing starts. the rollup addon glob ignores `src/addons/**/samples/**` so those bundles aren't emitted, and nine pages were still pointing at `build/addons/**/samples/**/main.js`. they were the only files left on that path.

second thing underneath it: `NetSample.init()` still started with `xb.core.input.raycaster.params.Line.threshold`, and `Input.raycaster` isn't part of the input API anymore. throws on the first line, `ScriptsManager` marks the script failed, so nothing ever got as far as `joinRoom`. fixing the 404 alone wouldn't have done it.

samples now sit next to their html under `demos/netblocks/src/` and build through the existing demos glob, same as `demos/animalattack` and `samples/advanced/drone`. that glob compiles each file on its own and keeps imports external, hence the `.js` on relative imports and `xrblocks/addons/...` instead of bare `netblocks` / `lipsync`, which resolves for tsc and the importmap both.

fixed the edge lines while in there. `objects.ts` set a flat `edges.pointerEvents`, which typechecks through the uikit `Object3D` augmentation but nothing reads it, the hit pipeline looks at `object.xb.pointerEvents`. integration wasn't marking them at all. both set `edges.xb` now so `EdgesGeometry` stops eating hits. related to #569.

dropped `lipsync/samples/puppet` instead of moving it, `samples/avatar_lab/lipsync_puppet` already covers it and is what the docs page links to.

`roomCode.test.ts` moved with its source so vitest picks up demos tests, and the demos glob skips test files now instead of emitting one as a demo bundle.

all nine pages load their bundle locally and join a room. integration and chat verified running end to end, 4/4 edge lines marked (was 0/4). build clean under `--failAfterWarnings`, lint clean, test totals unchanged.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

- **Simulator Recording**: verified in the desktop simulator against a local build, no recording attached
- **Device Recording**: not tested on device

## Checklist

- [x] **Tested in simulator & device**: verified in the desktop simulator; not run on hardware
- [x] **Large Assets ($\ge$ 1MB)**: none added
- [x] **SDK Dynamic Dependencies**: no new dependencies
- [x] **Security**: no keys or secrets


dli7319:
Thanks.

Where did you move "xb.core.input.raycaster.params.Line.threshold = 0.0001;" to?
I'm gonna set this in XR Blocks itself.

dli7319 (approved):

salmanmkc:
haha, we crossed by about 5 seconds, you asking here and me answering over on #569.

tl;dr for anyone landing on this later: it's deleted rather than moved. the edge lines opt out instead, `edges.xb = {pointerEvents: 'none'}`. setting the default in XR Blocks itself sounds like the better fix for the generated-app case.

PR #587, opened 2026-09-14
test: use jsdom storage in test workers

salmanmkc:
## Description

running the suite on Node 26 leaves 15 tests without jsdom's `localStorage`. Node defines the property itself now, so vitest keeps the native global instead of copying jsdom's Storage over. the same tests pass on Node 24, which is what CI uses. matches [vitest-dev/vitest#10867](https://github.com/vitest-dev/vitest/issues/10867).

turns off native Web Storage in the test workers, and only when the runtime accepts the flag, so jsdom's real Storage comes through. the fallback test gets instance and window-identity assertions so a mock can't satisfy it. no mocks, no dependency changes. test configuration and assertions only, no shipped SDK code changes.

rebased onto `a55aae65` after #585, then reran both baselines and the patched suite. 1721 tests in every run:

| runtime | before | after |
|---|---|---|
| Node 26.7.0 | 1698 passed, 15 failed, 8 skipped | 1713 passed, 0 failed, 8 skipped |
| Node 24.21.0 | 1713 passed, 0 failed, 8 skipped | 1713 passed, 0 failed, 8 skipped |

 the existing load-sensitive simulator-init test in `Core.test.ts` appeared in earlier full runs on both runtimes, but not in these post-rebase runs. left for a separate follow-up. all 36 tests in the three affected files pass on Node 22, 24 and 26, and the flag exits clean when passed straight to Node 22.23.2 and 24.21.0. `npm run lint` and `npm run format:check` pass.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

- **Simulator Recording**: N/A, test environment only.
- **Device Recording**: N/A, test environment only.

## Checklist

- [ ] **Tested in simulator & device**: Verified functionality in desktop simulator and/or physical hardware (where applicable). N/A, test environment only.
- [ ] **Large Assets ($\ge$ 1MB)**: Submitted separately to [xrblocks/proprietary-assets](https://github.com/xrblocks/proprietary-assets) via jsdelivr CDN. N/A, no assets.
- [ ] **SDK Dynamic Dependencies**: All new SDK dependencies are dynamically loaded at runtime. N/A, no dependency changes.


dli7319:
Thanks!

PR #593, opened 2026-09-15
fix(depth): convert view UVs to top-origin before the depth transform

salmanmkc:
## Description

`getDepth` and `getVertex` applied `normDepthBufferFromNormView` to the view UVs and flipped V afterwards. `DepthMesh` flips V first. both read the same buffer, so they only agreed when the transform commuted with that flip, which includes the identity that aligned devices report. that's why nothing caught it.

the [webxr algorithm](https://immersive-web.github.io/depth-sensing/#obtain-depth-at-coordinates) takes top-origin normalized view coords, applies the transform, then scales straight into the buffer with no second flip. chromium's `xr_cpu_depth_information.cc` does the same. so `DepthMesh` had the order right and `Depth` didn't.

pulled the conversion into one helper so `getDepth` and `getVertex` can't drift apart again.

the helper hands back top-origin buffer coords, so `getVertex` now converts back to upward-positive clip Y when it reconstructs the position. that's bookkeeping that falls out of the ordering fix, not a second bug: the old code sampled `1 - v` and built clip Y from `v`, which was already self-consistent.

the one in-repo caller is `samples/advanced/rain`, which passes `(ndc + 1) / 2`, so bottom-origin. that contract is now on the TSDoc of both public methods and in the manual, neither of which stated an origin before. the missing origin is basically how the two paths drifted.

two things I left alone on purpose, both pre-existing:

- `Depth` scales by `width`/`height` and `DepthMesh` by `width-1`/`height-1`, so they still land up to a pixel apart on interior samples. that's a sampling-policy question, not an ordering one, and it wants its own change.
- `getVertex` unprojects buffer coords through the depth camera's projection, which isn't right when the buffer is rotated relative to the camera. no callers today.

## Type of Change

- [x] Bug fix
- [ ] New demo or sample
- [ ] Documentation update

- **Simulator Recording**: n/a, the simulator reports an identity transform, where this is a no-op
- **Device Recording**: n/a, but measured rather than assumed. a diagnostic page on a Quest 3 reports `normDepthBufferFromNormView` as exact identity, max delta 0 against the identity matrix, on a 320x320 unsigned-short GPU-optimized buffer. so this is a no-op there and no recording can show a difference. worth stating plainly: no device I can test reports a non-identity transform, so this is a latent spec-correctness fix, not a fix for something users are hitting today. Here's a screenshot from the quest:
quest test

## Checklist

- [x] **Tested in simulator & device**: covered by unit tests rather than a recording, since the bug is invisible wherever the transform is identity.

  the main one drives a real `Depth` instance against a transcription of the spec algorithm, 4000 random UVs across five transforms, with the rounding policy held constant so only the flip order is under test. it also runs the previous ordering and asserts it *disagrees* wherever the transform doesn't commute with the flip, so the test can't quietly become a tautology.

  result: the new code matches the spec on all five. the old code matches on identity and a pure X mirror (both commute, which is why aligned devices saw nothing wrong) and diverges on rot90, a UV swap, and a non-commuting scale+offset. on a 160x120 buffer with a 90-degree transform the old ordering sampled a mean of 95px away; the new one is within 1px of `DepthMesh`, that 1px being the `width-1` difference noted above.

  full suite 1944 passing, lint and prettier clean.

  also ran the same oracle in real chromium on webgl2 (ANGLE) rather than jsdom, sampling the actual texture path, 4096 uvs per transform, identical harness on both builds. wrong samples before -> after: identity 0 -> 0, rot90 384 -> 0, swapUV 384 -> 0, scaleShift 465 -> 0. the identity row is the interesting one, it's the reason this sat here unnoticed: under identity both orderings are bit-identical, so there is no behaviour change to regress on aligned hardware.
- [x] **Large Assets ($\ge$ 1MB)**: n/a
- [x] **SDK Dynamic Dependencies**: n/a
- [x] **Security**: no keys or secrets


dli7319:
I think android phones return a non-identity transform. Will check on a phone.

dli7319:
Tested on my pixel 7 pro.

salmanmkc:
> Tested on my pixel 7 pro.

Awesome thanks, and thanks for getting through all these PRs!

PR #626, opened 2026-09-23
feat(ui): resize cards from their edge corners

### LINKED ISSUE #621: FR: Add resizing from the corners of panels
Similar to Android XR and Quest panel resizing.

salmanmkc:
## Description

closes #621. cards can now be resized by grabbing a corner of their edge band, like Android XR and Quest panels. resizing changes `size` and reflows the layout through uikit/yoga, it doesn't scale the content.

**resizing**

- grab a corner of the edge band to resize, the sides still move the card (same split as Quest)
- grows around the center by default, like Quest and Android XR. `resize: {anchor: 'opposite'}` pins the far corner instead
- `minSize` / `maxSize` in meters (no max by default)
- `preserveAspectRatio: true` keeps the card's proportions, off by default like both platforms
- never narrower than its content allows (longest word, a row of buttons), measured with yoga when the drag starts. set `minSize.width` to turn that off
- never shorter than its content at the current width, so making a card narrower makes it taller as text wraps. set `minSize.height` to turn that off, e.g. when the card scrolls its own content
- `height: 'auto'` cards only switch to a fixed height once you actually resize them, a press without a drag leaves them alone
- works from a controller ray, a hand pinch-ray, or pinching the corner directly with a tracked hand
- the hovered corner lights up with a thicker arc
- scripts get `onObjectManipulate` events with `action: 'resize'` and the new `width`, `height`, and `position`
- an edge can be resize-only with `manipulation: {actions: {resize: true}}`

**moving** (only with `manipulation: true`, an app's own `translate` options don't pick these up)

- distance scaling follows Android XR's spatial UI guide: same apparent size up to 1.75 m, then scale grows at 0.5 m per meter
- controller thumbstick pushes the card away (forward) or pulls it closer (back) while dragging. only `xr-standard` gamepads, so the simulator's right stick still just looks around
- every move stays within 0.75 to 5 m of the viewer (Android XR's panel limits). a card that starts outside can still move, just not farther out
- opt in from your own options with `translate: {scaleWithDistance, pushPull, minDistance, maxDistance}`
- two-hand and scroll-wheel scale work the same as before

**sample**

- the spatial_ui Explore card now puts its body in a `UIScrollView` with a `minSize.height`, so it can shrink past its content and scroll

no resize placeholder for now, live reflow felt fine on device. can add one later if heavier cards struggle.

## Type of Change

- [ ] Bug fix
- [ ] New demo or sample
- [x] Documentation update

- **Simulator Recording**: screenshots from the desktop simulator below. I've tested on a Meta Quest and controller to move panels back and forth with the joystick works like on meta quest native panels.
- **Device Recording**: 

| before | Explore resized larger |
| --- | --- |
| before | Explore card resized larger |
| **Explore shrunk, body scrolls** | **Tune narrower, grows taller to fit** |
| Explore card shrunk below its content with a scrolling body | Tune card narrowed, text wraps and the card grows taller |
| **Tune and Spatial dragged as narrow as they go** | |
| Tune and Spatial cards at their narrowest, content wraps without overflowing | |

## Checklist


dli7319:
Thank you!

Can you also change the text on the card on the right to feature resizing instructions instead of scaling instructions?

salmanmkc:
> Thank you!
> 
> Can you also change the text on the card on the right to feature resizing instructions instead of scaling instructions?

yup done!

dli7319:
I tested it on Galaxy XR and it's working very well!
I'm gonna make a few minor fixes before I merge it.

dli7319 (approved):

